PR #5, opened 2023-09-08
Add color module

cgohlke:
This PR adds a new color module to the PhasorPy package. 

For now it provides two color palettes and a ``wavelength2rgb`` function, which maps visible wavelengths to approximate sRGB color components. It can, for example, be used to colorize hyperspectral image stacks for visualization.

More specialty palettes and functions for compositing phasor histograms and intensity images may be added to the module.


bruno-pannunzio (approved):
Great PR, I looked through it and I think it will be useful. I used some colormaps from matlotlib for other uses and I think this is a better approach. I will try to think how to use something like this for mapping the phasor circle to a colormap.

Issue #10, opened 2023-10-16
Support PicoQuant PT3 and PTU files


cgohlke:
Dr Evangelos Sisamakis, Product Manager Microscopy at [PicoQuant GmbH](https://www.picoquant.com), kindly provided some resources, which may facilitate implementing a file reader for the PicoQuant PT3 and PTU file formats.

The [demo workspace dataset](https://www.picoquant.com/dl_software/SPT64/Demo_workspace_SPT64.zip) contains data files used in the [SymPhoTime Analysis tutorials](https://www.tcspc.com/doku.php/howto:start).

The [SPT 64 tutorials dataset](https://figshare.com/s/4957fcfa684daef86c23) contains data files for the [Lifetime-Fitting Using the FLIM Analysis](https://www.tcspc.com/doku.php/howto:lifetime-fitting_using_the_flim_script) tutorial.

The [Luminosa software](https://www.picoquant.com/dl_software/Luminosa/LuminosaCompactInstaller.zip) can be used freely to open data files taken with a Luminosa system.

A few Python scripts (but no ready to use library) are available to read PTU and PT3 files:

- [Python/Read_PTU.py](https://github.com/PicoQuant/PicoQuant-Time-Tagged-File-Format-Demos/blob/master/PTU/Python/Read_PTU.py): official reference implementation

- [readPTU_FLIM.py](https://github.com/SumeetRohilla/readPTU_FLIM/blob/master/readPTU_FLIM.py)

- [picoquantio](https://github.com/tsbischof/picoquantio): unreleased Python package

- [snAPI wrapper](https://github.com/PicoQuant/snAPI): for time traces and correlation curves, not images

The data T3-mode formats are described at https://www.tcspc.com/doku.php/glossary:t3-mode.

cgohlke:
I [released](https://pypi.org/project/ptufile/) a first version of the [ptufile](https://github.com/cgohlke/ptufile) library.

Mariochem92:
> I [released](https://pypi.org/project/ptufile/) a first version of the [ptufile](https://github.com/cgohlke/ptufile) library.

Thank you very much @cgohlke I tested this and it works like a charm. Do you think it would be possible to make it straightforward to export the decay curves from each pixel?

cgohlke:
Closing this issue. A reader for PTU files was added in #25. Support for PT3 files would need to be added to the ptufile library first, which seems unlikely anytime soon given that the format is deprecated for ~10 years and no formal format specification is available.

[...]

PR #25, opened 2024-01-04
Add read functions for PTU, FBD, and FLIF files

cgohlke:
## Description

This PR adds read functions for PTU, FBD, and FLIF files to the `phasorpy.io` module. It also fixes the IFLI spectral axes code and makes `read_sdt` accept SDT FCS data files.

The [ptufile](https://github.com/cgohlke/ptufile) library is added as an optional dependency required for reading PTU files.

Partly addresses issue #10.

## Release note

```release-note
Add read functions for PTU, FBD, and FLIF files
```

## Checklist

### DIFF
--- a/src/phasorpy/io.py
+++ b/src/phasorpy/io.py
@@ -723,9 +723,10 @@ def read_fbd(
     filename: str | PathLike[Any],
     /,
     *,
+    frame: int | None = None,
     channel: int | None = None,
+    keepdims: bool = True,
     laser_factor: float = -1.0,
-    integrate_frames: int = 1,
 ) -> DataArray:
     """Return frequency-domain image and metadata from FLIMbox FBD file.
 
@@ -743,17 +744,21 @@ def read_fbd(
     ----------
     filename : str or Path
         Name of FLIMbox FBD file to read.
+    frame : int
+        If None (default), return all frames.
+        If < 0, integrate time axis, else return specified frame.
+    channel : int
+        If None (default), return all channels, else return specified channel.
+    keepdims :
+        If true (default), reduced axes are left as size-one dimension.
     laser_factor : float
         Factor to correct dwell_time/laser_frequency.
-    integrate_frames : int
-        Specifies which frames to sum. By default, all frames are summed
-        into one. If 0, no frames are summed.
 
     Returns
     -------
     xarray.DataArray
         Frequency-domain image histogram with :ref:`axes codes <axes>`
-        ``'CYXH'`` and type ``uint16``:
+        ``'TCYXH'`` and type ``uint16``:
 
         - ``coords['H']``: phases in radians.
         - ``attrs['frequency']``: repetition frequency in MHz.
@@ -771,7 +776,7 @@ def read_fbd(
     >>> data.dtype  # doctest: +SKIP
     dtype('uint16')
     >>> data.shape  # doctest: +SKIP
-    (1, 2, 256, 256, 64)
+    (9, 2, 256, 256, 64)
     >>> data.dims  # doctest: +SKIP
     ('T', 'C', 'Y', 'X', 'H')
     >>> data.coords['H'].data  # doctest: +SKIP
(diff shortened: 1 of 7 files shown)


src/phasorpy/io.py
```diff
@@ -522,6 +540,266 @@ def read_sdt(
     return DataArray(data, **metadata)
 
 
+def read_ptu(
```
bruno-pannunzio, line 543: Tested the function with a couple of files and worked as expected. The output was the same as before with using the lfdfiles.

src/phasorpy/io.py
```diff
@@ -522,6 +540,266 @@ def read_sdt(
...
+    filename: str | PathLike[Any],
+    /,
+    selection: Sequence[int | slice | EllipsisType | None] | None = None,
```
bruno-pannunzio, line 546: I didn't test the selection argument because I had some trouble understanding what it is and the different options.

The other arguments were tested and worked as expected

src/phasorpy/io.py
```diff
@@ -522,6 +540,266 @@ def read_sdt(
...
+
+    Returns
+    -------
+    xarray.DataArray
+        Decoded TTTR T3 records as up to 5-dimensional image array
+        with :ref:`axes codes <axes>` ``'TYXCH'`` and type specified
+        in ``dtype``:
+
+        - ``coords['H']``: times of the histogram bins.
+        - ``attrs['frequency']``: repetition frequency in MHz.
```
bruno-pannunzio, line 596: The frequency returned in the files I tested was not correct, but I guess that is already wrong in the metadata of the file

src/phasorpy/io.py
```diff
@@ -522,6 +540,266 @@ def read_sdt(
...
+            keepdims=keepdims,
+            asxarray=True,
+        )
+
+    assert isinstance(data, DataArray)
+    data.attrs['frequency'] *= 1e-6  # MHz
+    return data
+
+
+def read_flif(
```
bruno-pannunzio, line 645: Coudn't test this function because I haveno flif files

src/phasorpy/io.py
```diff
@@ -522,6 +540,266 @@ def read_sdt(
...
+        attrs['frequency'] = float(flif.header.frequency)
+        attrs['ref_phase'] = float(flif.header.measured_phase)
+        attrs['ref_mod'] = float(flif.header.measured_mod)
+        attrs['ref_tauphase'] = float(flif.header.ref_tauphase)
+        attrs['ref_taumod'] = float(flif.header.ref_taumod)
+
+    return DataArray(data, **metadata)
+
+
+def read_fbd(
```
bruno-pannunzio, line 722: This function also worked perfectly. I tested with different data and returned the expected data. I think it would be great if it can also be implemented the 'frame', 'channel' and 'keepdims' arguments like in the read_ptu function so that the returned image already has only three dimensions.

bruno-pannunzio (approved):
Thanks for implementing this functions Christoph. I tested read_ptu and read_fbd and they work perfectly. I added some comments because I didn't understand a couple of arguments and also gave some feedback on implementing something for when reading all files that can be very useful so that it can be possible to trim the dimensions.

src/phasorpy/io.py
cgohlke: The `selection` parameter allows numpy style indexing with slices, integers, and ellipsis, except that step sizes mean integration. See for example https://github.com/cgohlke/ptufile/blob/3a2741fd3b21101d91f55ace8c55de0f5d781135/ptufile/ptufile.py#L172-L178 and https://github.com/cgohlke/ptufile/blob/3a2741fd3b21101d91f55ace8c55de0f5d781135/tests/test_ptufile.py#L451-L461. It's an advanced option that would allow, for example, to define the maximum number of bins returned when used with `trimdims`.

src/phasorpy/io.py
cgohlke: What value was returned and what did you expect? The frequency is calculated at https://github.com/cgohlke/ptufile/blob/3a2741fd3b21101d91f55ace8c55de0f5d781135/ptufile/ptufile.py#L1170-L1175. It applies to the maximum number of bins, not the actual number of bins returned. Are the `coords['H']` values correct?

src/phasorpy/io.py
cgohlke: There's a test file on Zenodo. This format is of no practical relevance. I just implemented it to get access to some old homodyne frequency-domain data.

src/phasorpy/io.py
bruno-pannunzio: The frequency returned is 2.517700195304632 and I was told by the people who acquired the image that the frequency was 80 MHz. I don't know if the values in coords['H'], I will try to see if the problem lies there.

cgohlke: I see. The "frequency" here is not the laser pulse rate but is calculated from the TCSPC resolution and the number of bins of the hardware. It's probably not useful and better to return the value of the `TTResult_SyncRate` PTU tag instead, or the inverse of `MeasDesc_GlobalResolution`. I'm not sure what to do if the time of the last bin with non-zero photons exceeds the duration between sync events...

cgohlke: I changed the frequency to `TTResult_SyncRate`. Can you check again with your files?

src/phasorpy/io.py
cgohlke: I added/changed the `frame`, `channel`, and `keepdims` parameters to the `read_fbd` function to make it more alike to `read_ptu`.

src/phasorpy/io.py
bruno-pannunzio: Now it shows the correct frequency (78.02 MHz). Thanks for improving this.

src/phasorpy/io.py
bruno-pannunzio: Thanks for adding this options, I tested them in read_fbd and they work great and helps to speed up the process of importing the data intro the correct format.

PR #20, opened 2023-12-12
Add phasor_from_signal function

schutyb:
## Description

... This PR adds a phasor module in phasorpy that compute the phasor transform. 

## Release note

1. Add phasorpy.phasor 
2. Add a test file for the phasor function 

```release-note
...
```

## Checklist

- [ ] The pull request title, summary, and description are concise.
- [ ] Related issues are linked in the description.
- [ ] New dependencies are explained.
- [ ] The source code and documentation can be distributed under the [MIT license](https://www.phasorpy.org/stable/license.html).
- [ ] The source code adheres to [code standards](https://www.phasorpy.org/stable/contributing.html#code-standards).
- [ ] New classes, functions, and features are thoroughly [tested](https://www.phasorpy.org/stable/contributing.html#tests).
- [ ] New, user-facing classes, functions, and features are [documented](https://www.phasorpy.org/stable/contributing.html#documentation).
- [ ] New features are covered in tutorials.
- [ ] No files other than source code, documentation, and project settings are added to the repository.

PR #13, opened 2023-11-01
Lifetime module branch

Mariochem92:
Tackling issue #6 : This update facilitates phasor coordinate calculations for frequency-domain signals, conversion to phase, and effective single lifetime and mixture management. I am considering  introducing spatial filtering and a couple of extra features but that also depends on the overall project structure you have in mind.Future enhancements, are open for community contributions, as I'm not well-versed in FRET or time-domain conversion.

PR #32, opened 2024-02-04
Various additions and improvements to the phasor module

cgohlke:
## Description

This PR proposes various additions and improvements to the `phasorpy.phasor` module and the build infrastructure:

- add [Cython](https://cython.readthedocs.io/en/stable/index.html) as a build dependency
- add workflow to build binary wheels with [cibuildwheel](https://github.com/pypa/cibuildwheel)
- complete `phasor_from_lifetime`, including speedups written in Cython
- add simple tutorial for `phasor_from_lifetime`
- add `phasor_from_signal_f1` function, including speedups in Cython. Alternative to #20 `phasor_from_signal`
- add `phasor_from_polar` function. Replaces #13 `phasor_coordinates`
- add `phasor_semicircle` helper function
- use [Sphinx `:math:`](https://www.sphinx-doc.org/en/master/usage/extensions/math.html) for formulas
- fix minor docstring issues

The following are being postponed to a later PR:

- update "Publish to PyPI" workflow to cibuildwheel.
- implement `phasor_from_apparent_lifetime` and `phasor_to_apparent_lifetime`.

A Python compatible C compiler is now required for developing the PhasorPy library.

## Release note

```release-note
Various additions and improvements to the phasor module
```

## Checklist


bruno-pannunzio:
Hi @cgohlke, thanks for improving the docstrings of the functions that had some errors. I don't think this overlaps with anything I am working on so no problem with that. I like the idea very much of starting to implement some thing in cython to speed up the processing, maybe we can talk about it today in the meeting.

I also believe that the tests should be improved as there are many cases that are not testes, we can focus on improving this on the following weeks.

Regarding the `phasor_from_signal` implementation, I believe Bruno has been researching on ways to improve it and also to accept different fft functions, so maybe we can also discuss today this aspect.

cgohlke:
Hi @bruno-pannunzio . Thanks for the feedback.

OK, I will remove `phasor_from_signal` from this PR (assuming #20 is imminent) but leave `phasor_from_signal_f1`.

Besides the tests marked with `TODO:`, any other test cases need to be improved in this PR?

bruno-pannunzio:
I think those are ok. I think they also apply to the tests from my previous PR because not all functions cover all those test cases you added to the TODO. So I will improve those tests also when possible.

bruno-pannunzio:
You are right, maybe its something that we can discuss if it make sense or if it can lead to misunderstandings. I will get feedback on this topic.

cgohlke:
Interesting benchmark results: the `phasor_from_signal_f1` function is 10 to 100 times faster than the `phasor_from_signal` reference numpy implementation on a 256x256x256 uint8 stack ([lifetime_cat.tif](https://github.com/zoccoler/napari-flim-phasor-plotter/tree/main/src/napari_flim_phasor_plotter/data)). Maybe this computation is extremely memory limited or numpy converts the stack to double precision?

```
timeit.timeit('phasor_from_signal(stack, axis=0)', number=10, globals=globals())
6.033
timeit.timeit('phasor_from_signal_f1(stack, axis=0)', number=10, globals=globals())
0.524
timeit.timeit('phasor_from_signal_f1(stack, axis=0, numthreads=28)', number=10, globals=globals())
0.063
```

cgohlke:
Hi @bruno-pannunzio. I think this is ready for review.

You can download the docs, including formulas and tutorial as a build artifact at https://github.com/phasorpy/phasorpy/actions/runs/7825054348?pr=32

Never mind the long `phasor_plot` function in the tutorial. The plan is to move it to a `phasorpy.plot` module.

cgohlke:
I see the following issues with the `phasor_from_signal_f1` function:

1. A copy of the input signal array is made if it is not C contiguous. That is necessary because the array is reshaped to a 3D memoryview. This could be avoided by using the [numpy iterator C API](https://numpy.org/devdocs/reference/c-api/iterator.html), but the implementation would be less straightforward/readable/maintainable and not compatible with OpenMP (I think).
2. On macOS, Xcode does not include OpenMP libraries. There's a [workaround using clang](https://mac.r-project.org/openmp), but that could break anytime and might conflict with workarounds used in other Python libraries.
3. Related to that, a slower than necessary code path is used when setting `num_threads` > 1 on builds without OpenMP.
4. Only the phasor coordinates of the fundamental frequency is calculated.

bruno-pannunzio:
Hi Christoph, sorry I didn't reply before, I was on a trip until yesterday. I will go through your comments and will get back to you on the aspects you pointed out.

bruno-pannunzio:
Hi @cgohlke, I tested the `phasor_from_f1` and `phasor_from_lifetime` functions and both work great. I tried all the pipeline for calibration with real images (fbd and ptu) and it works as expected using both. Also I timed the phasor calculation from signal of ptu files which are quite large. Your implementation with cython is much faster than using numpy fft. Using `phasor_from_signal_f1` took just 2.8 seconds, while the numpy fft implementation took 13.5 seconds, so this is a great improvement.

I also tested the `num_threads` parameters but had not much of an impact with different numbers tested. Maybe as I am running with macOS, this is not compatible as you stated before. 

The tutorial looks really good and easy to understand. The only thing I see is a bit distracting is the code for `phasor_plot` at the beginning, but as you said before, when that is implemented in `phasorpy.plot`, I'm guessing it will disappear from the tutorial and it will be cleaner.

src/phasorpy/phasor.py
```diff
@@ -34,23 +35,216 @@
...
+    # pure numpy implementation for reference:
+    # shape = [1] * signal.ndim
+    # shape[axis] = sample_phase.size
+    # sample_phase = sample_phase.reshape(shape)  # make broadcastable
+    # mean = numpy.mean(signal, axis=axis)
+    # real = numpy.mean(signal * numpy.cos(sample_phase), axis=axis)
+    # real /= mean
+    # imag = numpy.mean(signal * numpy.sin(sample_phase), axis=axis)
+    # imag /= mean
+    # return mean, real, imag
```
bruno-pannunzio, line 172: Should we keep this commented implementation?

src/phasorpy/phasor.py
```diff
@@ -34,23 +35,216 @@
...
+    _phasor_from_signal(phasor, signal, sincos, num_threads)
+
+    # restore original shape
+    shape = shape0 + shape1
+    mean = phasor[0].reshape(shape)
+    real = phasor[1].reshape(shape)
+    imag = phasor[2].reshape(shape)
+    if shape:
+        return mean, real, imag
+    return mean.item(), real.item(), real.item()
```
bruno-pannunzio, line 195: In this case the real component is returned as if it was the imaginary?

src/phasorpy/phasor.py
```diff
@@ -34,23 +35,216 @@
...
+    ...     numpy.cos(sample_phase - 0.78539816) * 2 *  0.70710678 + 1
+    ... )
+    >>> phasor_from_signal_f1(
+    ...    signal, sample_phase=sample_phase
+    ... )  # doctest: +NUMBER
+    (1.1, 0.5, 0.5)
+
+    """
+    signal = numpy.array(signal, order='C', ndmin=1, copy=False)
+    # assert signal.flags['C_CONTIGUOUS']
```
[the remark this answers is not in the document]
cgohlke: This should be taken care of by `order='C'` in the line before. I just kept this as a comment/reminder in case the previous line changes. I will double-check that there's a test case.

src/phasorpy/phasor.py
cgohlke: I was thinking to move it into `test_phasor.py` or `phasorpy.testing_utils.py`, but then the test function would need to be tested too. Also, the plan is to test `phasor_from_signal_f1` against `phasor_from_signal` once implemented, or to merge those functions into one...

src/phasorpy/phasor.py
cgohlke: Ouch. That should have been caught by the tests.

cgohlke:
Thanks for the feedback.

> I also tested the num_threads parameters but had not much of an impact with different numbers tested. Maybe as I am running with macOS, this is not compatible as you stated before.

Yes, unfortunately there's no official support for OpenMP on macOS/Xcode. It's unbelievable. I applied the workaround in the `lfdfiles` wheels, which ship with their own copy of OpenMP. I 'm not sure it's the right solution for this project.

> The only thing I see is a bit distracting is the code for phasor_plot at the beginning, but as you said before, when that is implemented in phasorpy.plot, I'm guessing it will disappear from the tutorial and it will be cleaner.

Exactly, that function will be moved to `phasorpy.plot`. I have done some work on that. It will be in another PR.

src/phasorpy/phasor.py
bruno-pannunzio: You are right, it can be a good a idea to keep that implementation somewhere to compare. The proposed `phasor_from_signal` is also different from that implementation, so maybe we can keep and compare the three of them?

src/phasorpy/phasor.py
cgohlke: Thanks again for catching this. It is fixed and tested now.

src/phasorpy/phasor.py
cgohlke: For now there is no need for this in the tests, which use known values. So I think to leave it as a comment in this PR. Another option would be to implement it as a reference in a tutorial/benchmark that compares various implementations.

src/phasorpy/phasor.py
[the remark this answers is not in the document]
cgohlke: I added a test for a non-contiguous signal array.

src/phasorpy/phasor.py
```diff
@@ -34,23 +35,215 @@
...
+    real = phasor[1].reshape(shape)
+    imag = phasor[2].reshape(shape)
+    if shape:
+        return mean, real, imag
+    return mean.item(), real.item(), imag.item()
+
+
+def phasor_semicircle(
+    samples: int = 33, /
+) -> tuple[NDArray[numpy.float64], NDArray[numpy.float64]]:
```
cgohlke, line 199: How about removing the arbitrary default?
```Python
def phasor_semicircle(
    samples: int, /
) -> ...
```

src/phasorpy/phasor.py
```diff
@@ -34,23 +35,215 @@
...
 import numpy
 
+from ._phasor import _phasor_from_lifetime, _phasor_from_signal
+
+
+def phasor_from_signal_f1(
+    signal: ArrayLike,
+    /,
+    *,
+    axis: int = 0,
```
cgohlke, line 67: Just like to mention that this default does not work in many cases, like histograms returned from PTU files. Should we change it to `-1`, to match the default in numpy FFT? `-1` will also b the fastest to compute. On the other hand, `-1` will not work by default with many spectral signals. Thoughts?

src/phasorpy/phasor.py
```diff
@@ -282,47 +482,154 @@ def phasor_to_polar(
...
+    frequency: ArrayLike,
     lifetime: ArrayLike,
     /,
     fraction: ArrayLike | None = None,
     *,
-    is_preexp: bool = False,
+    preexponential: bool = False,
+    unit_conversion: float = 1e-3,
+    squeeze: bool = True,
 ) -> tuple[NDArray[numpy.float64], NDArray[numpy.float64]]:
```
cgohlke, line 541: How about allowing frequency and lifetime to be named arguments? That could make the tutorial clearer.
```Python
def phasor_from_lifetime(
    frequency: ArrayLike,
    lifetime: ArrayLike,
    fraction: ArrayLike | None = None,
    *,
    is_preexp: bool = False,
) ->
```

src/phasorpy/phasor.py
bruno-pannunzio: I thought about that when testing with real data, because I always need to perform some sort of transpose or indicate the axis after reading the files. I think it could be a good idea to leave this default axis as 0, and then maybe modify the read functions so that the returned data contains the information to perform the fft in the first dimension. I believe the spectral data like lsm returns in the first dimension and thats why we started with that workflow. But maybe we can discuss this in the next meeting if it is viable to unify this with spectral and flim.

src/phasorpy/phasor.py
bruno-pannunzio: I think it is good to have a default in this case, but I am not exactly sure what can be a better number. Also thinking about the use of this function, maybe you have better understanding on different use cases and what impact can have changing that value. 

Also, what could be the benefits of using a smaller/larger number as default?

src/phasorpy/phasor.py
bruno-pannunzio: Yes it can look more clear and easy to understand I think.

src/phasorpy/phasor.py
cgohlke: OK. Lets leave the default. The only use case for the function is (so far) for the `phasor_plot` function. If we don't see any other use, we can later consider to move this function.

src/phasorpy/phasor.py
cgohlke: Transposing the signal or modifying the read function to match `axis=0` is a bad idea. Ideally, for huge performance reasons, the `axis` should be contiguous, and that's why numpy defaults to `axis=-1` not only in the fft function. That's a really strong argument for using `axis=-1` as default. But if almost all of our use cases are using `axis=0`, that's a sensible default too.

In principle, `axis` can most often be determined from the xarray coordinates returned by the phasorpy.io read functions. I did not integrate it into this PR since it complicates things. Now that we have implemented the necessary functions, we were also discussing adding higher level functions to `phasorpy.io`, which directly return phasor coordinates from files, like `phasor_from_ptu`. Choosing the axis would be hidden from the user altogether.

src/phasorpy/phasor.py
cgohlke: Done

src/phasorpy/phasor.py
bruno-pannunzio: You are right, those higher level functions may help mitigate the issue I raised so I agree that default to -1 can be a good idea for the axis parameter.

cgohlke: I changed the default to `-1` for now. We can discuss further in #20.

cgohlke:
Hi @bruno-pannunzio Do you think this can be merged?

bruno-pannunzio:
@cgohlke I was just going to comment that I believe this can be merged and we can keep working/discussing the points were we can have more input.

bruno-pannunzio (approved):

cgohlke:
Btw, wheels for the main branch are available at https://github.com/phasorpy/phasorpy/actions/workflows/build_wheels.yml. Just follow the latest workflow run and go to artifacts, for example, https://github.com/phasorpy/phasorpy/actions/runs/7933102707#artifacts

PR #37, opened 2024-02-29
Higher level calibration function

bruno-pannunzio:
## Description
This PR renames the `phasor_calibrate` function to `phasor_transform` and adds the higher level function `phasor_calibrate` for easier calibration.

...

## Release note

```release-note
Add higher level function `phasor_calibrate` and rename old one to `phasor_transform`
```

## Checklist


bruno-pannunzio:
Hi @cgohlke, I created this PR so we can start discussing the approach of the higher level function `phasor_calibrate`. I haven't done the corresponding tests yet, but wanted to share beforehand to see if you consider the approach correct.

Particularly, I have questions with the optional arguments passed to other functions, such as `phasor_from_lifetime` and `phasor_center`. I left the option to pass them as kwargs, but I don't know if this is the best approach since they won't be directly visible when using `phasor_calibrate`. They would have to go to the documentation of those functions and I don't think is optimal. Maybe add something in the docstring of `phasor_calibrate` regarding those optional kwargs? What do you think.

src/phasorpy/phasor.py
```diff
@@ -387,34 +390,37 @@ def phasor_semicircle(
...
-
-def phasor_calibrate(
-    real: ArrayLike,
-    imag: ArrayLike,
-    phase0: ArrayLike = 0.0,
-    modulation0: ArrayLike = 1.0,
-    /,
-) -> tuple[NDArray[Any], NDArray[Any]]:
-    """Return calibrated/referenced phasor coordinates.
+def phasor_calibrate(real, imag, reference_real, reference_imag, frequency, lifetime, fraction, **kwargs):
```
cgohlke, line 393: This needs to be type annotated. `fraction` is optional.

src/phasorpy/phasor.py
```diff
@@ -387,34 +390,37 @@ def phasor_semicircle(
...
-        Imaginary component of phasor coordinates to calibrate.
-    phase0 : array_like, optional
-        Angular component of polar coordinates for calibration in radians.
-        Defaults to 0.0.
-    modulation0 : array_like, optional
-        Radial component of polar coordinates for calibration.
-        Defaults to 1.0.
+        Imaginary component of phasor coordinates to be calibrated.
+    reference_real : array_like
+        Real component of phasor coordinates from reference of known lifetime.
```
cgohlke, line 408: Maybe add that the reference phasor coordinates must be measured at the same instrument settings as the phasor coordinates to be calibrated.

src/phasorpy/phasor.py
```diff
@@ -387,34 +390,37 @@ def phasor_semicircle(
...
+        A scalar or one-dimensional sequence.
+    lifetime : array_like
+        Lifetime components in ns. See notes below for allowed dimensions.
+    fraction : array_like, optional
+        Fractional intensities or pre-exponential amplitudes of the lifetime
+        components. Fractions are normalized to sum to 1.
+        See notes below for allowed dimensions.
+    **kwargs
+        Additional keyword arguments passed to :py:func:`phasor_center`
+        and :py:func:`phasor_from_lifetime`,
```
cgohlke, line 423: I think it makes sense in this case to explicitly list the few parameters and repeat the definitions. This makes it easier for the user.

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+    return phasor_transform(re, im, phi_shift, mod_ratio)
+
+def phasor_transform(
+    real: ArrayLike,
+    imag: ArrayLike,
+    phase_shift: ArrayLike = 0.0,
+    modulation_ratio: ArrayLike = 1.0,
+    /,
+) -> tuple[NDArray[Any], NDArray[Any]]:
+    """Return transformed/rotated phasor coordinates.
```
cgohlke, line 475: -> "Return rotated and scaled phasor coordinates."

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+
+    This function is used for calibration purposes, but can also be used to
+    transform and/or rotate phasor coordinates. The transform is applied in
+    the polar system and returned as phasor coordinates.
```
cgohlke, line 479: -> "This function rotates and uniformly scales phasor coordinates around the origin.
It can be used, for example, to calibrate phasor coordinates."

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+        Angular component of polar coordinates for transformation/rotatio in radians.
+        Defaults to 0.0.
+    modulation_ratio : array_like, optional
+        Radial component of polar coordinates for transformation/rotatio.
+        Defaults to 1.0.
+
+    Returns
+    -------
+    real : ndarray
+        Transformed/rotated real component of phasor coordinates.
```
cgohlke, line 497: -> "Real component of rotated and scaled phasor coordinates."

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
     (array([-1.56, 1.934, 0.3279]), array([5.985, 10.6, 1.986]))
```
cgohlke, line 450: I would prefer to have a more realistic example, where the calibrated coordinates end up in the range 0..1.

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+    ValueError
+        The array shapes of `real` and `imag`, or `phase_shift` and `modulation_ratio`
+        do not match.
+
+    Examples
+    --------
+    Use scalar reference coordinates to transform/rotate phasor coordinates:
+
+    >>> phasor_trasnform([1.0, 2.0, 3.0], [4.0, 5.0, 6.0], 0.5, 0.2)
+    (array([-0.208, -0.1284, -0.04876]), array([0.798, 1.069, 1.341]))
```
cgohlke, line 512: `phasor_trasnform` ->  `phasor_transform`
Also here I would prefer a more realistic example.

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
 
     """
-    phi0 = numpy.asarray(phase0)
-    mod0 = numpy.asarray(modulation0)
-    if phi0.shape != mod0.shape:
-        raise ValueError(f'{phi0.shape=} != {mod0.shape=}')
+    re = numpy.asarray(real)
+    im = numpy.asarray(imag)
+    if re.shape != im.shape:
+        raise ValueError(f'{re.shape=} != {im.shape=}')
```
cgohlke, line 456: I think the error message should refer to the documented parameter names, `real` and `imag` in this case. The user does not know about `re` and `im`

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+    the polar system and returned as phasor coordinates.
+
+    Parameters
+    ----------
+    real : array_like
+        Real component of phasor coordinates to transform/rotate.
+    imag : array_like
+        Imaginary component of phasor coordinates to transform/rotate.
+    phase_shift : array_like, optional
+        Angular component of polar coordinates for transformation/rotatio in radians.
```
cgohlke, line 488: -> "Rotation angle in radians."

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+        Defaults to 0.0.
+    modulation_ratio : array_like, optional
+        Radial component of polar coordinates for transformation/rotatio.
```
cgohlke, line 491: -> "Uniform scale factor."

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
-        raise ValueError(f'{phi0.shape=} != {mod0.shape=}')
+    re = numpy.asarray(real)
+    im = numpy.asarray(imag)
+    if re.shape != im.shape:
+        raise ValueError(f'{re.shape=} != {im.shape=}')
+    ref_re = numpy.asarray(reference_real)
+    ref_im = numpy.asarray(reference_imag)
+    if ref_re.shape != ref_im.shape:
+        raise ValueError(f'{ref_re.shape=} != {ref_im.shape=}')
+    kwargs_center = {key: value for key, value in kwargs.items() if key in inspect.signature(phasor_center).parameters}
```
cgohlke, line 461: I think `inspect.signature` is not the right approach here. Better not to use `kwargs` in this case. See comment above.

src/phasorpy/phasor.py
```diff
@@ -444,18 +450,87 @@ def phasor_calibrate(
...
+    """Return transformed/rotated phasor coordinates.
+
+    This function is used for calibration purposes, but can also be used to
+    transform and/or rotate phasor coordinates. The transform is applied in
+    the polar system and returned as phasor coordinates.
+
+    Parameters
+    ----------
+    real : array_like
+        Real component of phasor coordinates to transform/rotate.
```
cgohlke, line 484: -> "Real component of phasor coordinates to transform."
Just refer to "transform", or alternatively "rotate and scale".

cgohlke:
I think the split into higher level `phasor_calibrate` and lower level `phasor_transform` functions make a lot of sense.

src/phasorpy/phasor.py
bruno-pannunzio: I agree, as they are not that many arguments, it won't be that long the list of arguments that it ends up being a mess. Maybe if for some reason the other functions are changed and more arguments are added we may need to revise this.

src/phasorpy/phasor.py
bruno-pannunzio: I didn't realize this. Thanks for pointing it out.

cgohlke:
> I have questions with the optional arguments passed to other functions, such as phasor_from_lifetime and phasor_center. I left the option to pass them as kwargs, but I don't know if this is the best approach

This is a recurring issue.
I think in this case of few parameters forwarded to functions within the same library, the use of kwargs should be avoided. Instead, explicitly list the parameters and repeat the documentation. That way the user does not have to look up documentation elsewhere and can use hints from the IDE. Also MyPy can verify the types.
In other cases, for example, when forwarding many possible parameters to matplotlib functions, it does not make sense to list and document those parameters again.

cgohlke:
>  also renamed the phase0 and modulation0 parameters to phase_shift and modulation_ratio as discussed before.

I can't seem to find a link to the previous discussion right now. I don't particular like the `phase0` and `modulation0` names, and `phase_shift` and `modulation_ratio` look nicer. However, I find those confusing, since (1) phase and modulation are practically always phase shifts and modulations relative to something, and (2) the phase shift and relative modulation in this case are very specific to calibration. In formulas (I am aware of), the phase shift and relative modulation used for calibration are usually denoted as `φ0` and `M0`, hence I chose `phase0` and `modulation0`. In the past I used `phase_zero` and `modulation_zero`. Maybe we can discuss and settle this in Monday's meeting?

bruno-pannunzio:
We mentioned it in #32, but I agree that it may be confusing using `phase_shift`and `modulation_ratio` because they are also used in other context, altough they look nicer and also are easier to understand what they mean. `phase_zero`and `modulation_zero` can be a good approach. I prefer them to `phase0`and `modulation0`.

Maybe if we add some math equations to the docstring it can be more visual what `phase_zero`and `modulation_zero`are and we can use those names.

I agree, lets discuss on monday.

cgohlke:
> Should it be the already calculated real and imag for the image to be calibrated and reference, as it as of now, or should it be the signals

I think the function should accept phasor coordinates, which are the central "data structure" of this library. Some file formats (REF, R64, IFLI) do not contain signals but phasor or polar coordinates. At some point we will add high level `io` functions that return phasor coordinates.

bruno-pannunzio:
That's a good Idea, having higher level `io` functions that call `phasor_calibrate` and maybe can have an option to already return the calibrated phasor coordinates if desired. I like that approach much better as this allows greater versatility in `phasor_calibrate`

cgohlke:
> higher level io functions that call phasor_calibrate and maybe can have an option to already return the calibrated phasor coordinates if desired.

Sorry, that's not really what I meant. Usually, signals of reference measurements are in separate files and that will require separate io and calibration steps. Some file formats like R64, IFLI, FLIF already contain referenced data or contain calibration information. Only in those cases the phasor coordinates will be returned calibrated. Otherwise they need to be calibrated, which I would keep separate, something like:

```Python
mean, real, imag, frequency = io.phasor_from_ptu(filename)
mean_ref, real_ref, imag_ref, _ = io.phasor_from_ptu(filename_ref)
real, imag = phasor_calibrate(real, imag, real_ref, imag_ref, frequency, lifetime, ...)
```

cgohlke:
Re `phasor_calibrate` documentation: how about adding a note or an example that it is a convenience function for:
```Python
phasor_transform(
    real,
    imag,
    *polar_from_reference_phasor(
        *phasor_center(
            reference_real, 
            reference_imag, 
            skip_axes, 
            method,
        ),
        *phasor_from_lifetime(
            frequency, 
            lifetime, 
            fration, 
            preexponential, 
            unit_conversion,
        ),
    ),
)
```

src/phasorpy/phasor.py
```diff
@@ -390,7 +388,22 @@ def phasor_semicircle(
...
-def phasor_calibrate(real, imag, reference_real, reference_imag, frequency, lifetime, fraction, **kwargs):
+
+def phasor_calibrate(
+    real: ArrayLike,
+    imag: ArrayLike,
+    reference_real: ArrayLike,
+    reference_imag: ArrayLike,
+    frequency: ArrayLike,
+    lifetime: ArrayLike,
+    /,
```
cgohlke, line 399: Allow `frequency` and `lifetime` to be named arguments, for readability?
```
    reference_imag: ArrayLike,
    /,
    frequency: ArrayLike,
    lifetime: ArrayLike,
    *,
```

src/phasorpy/phasor.py
```diff
@@ -390,7 +388,22 @@ def phasor_semicircle(
...
+    *,
+    fraction: ArrayLike | None = None,
+    preexponential: bool = False,
+    unit_conversion: float = 1e-3,
+    squeeze: bool = True,
```
cgohlke, line 404: I don't think `squeeze` is needed for this wrapper.

src/phasorpy/phasor.py
```diff
@@ -435,102 +464,139 @@ def phasor_calibrate(real, imag, reference_real, reference_imag, frequency, life
...
+            ),
+            *phasor_from_lifetime(
+                frequency,
+                lifetime,
+                fraction,
+                preexponential,
+                unit_conversion,
+            ),
+        ),
+    )
```
cgohlke, line 489: Use a block:
```
    This function is a convenience wrapper for the following operations::

        phasor_transform(
            real,
            imag,
            *polar_from_reference_phasor(
                *phasor_center(
                    reference_real,
                    reference_imag,
                    skip_axes,
                    method,
                ),
                *phasor_from_lifetime(
                    frequency,
                    lifetime,
                    fraction,
                    preexponential,
                    unit_conversion,
                ),
            ),
        )
```

src/phasorpy/phasor.py
```diff
@@ -435,102 +464,139 @@ def phasor_calibrate(real, imag, reference_real, reference_imag, frequency, life
...
         Defaults to 1.0.
 
     Returns
     -------
     real : ndarray
-        Transformed/rotated real component of phasor coordinates.
+        Real component of rotated and scaled phasor coordinates..
     imag : ndarray
-        Transformed/rotated imaginary component of phasor coordinates.
+        Imaginary component of rotated and scaled phasor coordinates..
```
cgohlke, line 560: `..` -> `.`

src/phasorpy/phasor.py
```diff
@@ -435,102 +464,139 @@ def phasor_calibrate(real, imag, reference_real, reference_imag, frequency, life
...
-
-    Use separate reference coordinates for each phasor coordinate:
-
     >>> phasor_calibrate(
-    ...     [1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [0.5, 0.2, 0.3], [1.5, 2.0, 0.3]
-    ... )
-    (array([-1.56, 1.934, 0.3279]), array([5.985, 10.6, 1.986]))
+    ...    [0.1, 0.2, 0.3], [0.4, 0.5, 0.6],
+    ...    [0.2, 0.3, 0.4], [0.5, 0.6, 0.7],
+    ...    80, 4
```
cgohlke, line 496: Use `frequency=80, lifetime=4` for clarity?

src/phasorpy/phasor.py
cgohlke: Or better, use a [Sphinx code_block](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/code_blocks.html).

cgohlke: Or better use a [Sphinx code-block](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block)

bruno-pannunzio:
@cgohlke I added the tests and modified the introduction tutorial. I think this is ready for review, but there may be some aspects to improve.

cgohlke:
I will review this later today.

src/phasorpy/phasor.py
```diff
@@ -391,78 +392,231 @@ def phasor_semicircle(
...
+    fraction: ArrayLike | None = None,
+    preexponential: bool = False,
+    unit_conversion: float = 1e-3,
+    method: Literal['mean', 'median'] = 'mean',
 ) -> tuple[NDArray[Any], NDArray[Any]]:
-    """Return calibrated/referenced phasor coordinates.
+    """
+    Return calibrated/referenced phasor coordinates.
 
     Calibration of phasor coordinates in fluorescence lifetime analysis is
```
cgohlke, line 409: -> `Calibration of phasor coordinates from time-resolved measurements is`

src/phasorpy/phasor.py
```diff
@@ -391,78 +392,231 @@ def phasor_semicircle(
...
+    reference_imag : array_like
+        Imaginary component of phasor coordinates from reference of known
+        lifetime.
+        Must be measured with the same instrument setting as the phasor
+        coordinates to be calibrated.
+    frequency : array_like
+        Laser pulse or modulation frequency in MHz.
+        A scalar or one-dimensional sequence.
+    lifetime : array_like
+        Lifetime components in ns. See notes below for allowed dimensions.
```
cgohlke, line 432: `See notes below for allowed dimensions.` -> `Must be scalar or one dimensional.`

src/phasorpy/phasor.py
```diff
@@ -391,78 +392,231 @@ def phasor_semicircle(
...
+    fraction : array_like, optional
+        Fractional intensities or pre-exponential amplitudes of the lifetime
+        components. Fractions are normalized to sum to 1.
+        See notes below for allowed dimensions.
```
cgohlke, line 436: `See notes below for allowed dimensions.` -> ```Must be same size as `lifetime`.```

tutorials/phasorpy_introduction.py
```diff
@@ -124,36 +124,14 @@
 _, measured_real, measured_imag = phasor_from_signal(reference_signal, axis=0)
```
cgohlke, line 124: `_, reference_real, reference_imag = ...`

tutorials/phasorpy_introduction.py
```diff
@@ -124,36 +124,14 @@
...
-from phasorpy.phasor import polar_from_reference_phasor
-
-phase0, modulation0 = polar_from_reference_phasor(
-    measured_real, measured_imag, known_real, known_imag
-)
-
-# %%
-# Finally, calibrate the raw phasor coordinates obtained from the signal:
+# Calibrate the raw phasor coordinates obtained from the signal to be
+# calibrated and the reference of known lifetime (Fluorescein, 4.2 ns):
```
cgohlke, line 128: `Calibrate the raw phasor coordinates with the reference coordinates of known lifetime (Fluorescein, 4.2 ns):`

Also, the above section must be adjusted since it no longer matches the shown workflow:
```
# Phasor coordinates from time-resolved measurements must be calibrated
# with the polar coordinates (phase and modulation) obtained from a
# reference standard of known lifetime, acquired with the same instrument
# settings.
```

cgohlke:
This is almost ready. Can you pull in the recent changes from main?

tutorials/phasorpy_introduction.py
bruno-pannunzio: What about:

```
# Phasor coordinates from time-resolved measurements must be calibrated
# with measurements obtained from a reference standard of known lifetime,
# acquired with the same instrument settings.
```

src/phasorpy/phasor.py
```diff
@@ -391,78 +392,231 @@ def phasor_semicircle(
...
+    coordinates:
+    .. math::
+        \phi' &= \phi + \phi_{0}
+
+    The scaling is performed in the polar system by taking the product of the
+    `modulation_zero` (:math:`M_{0}`) parameter with the calculated modulation
+    (:math:`M`) from the real and imaginary components of phasor coordinates:
+    .. math::
+        M' &= M \cdot M_{0}
+
```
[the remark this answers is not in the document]
bruno-pannunzio: I agree it's a little bit too verbose. I think what you wrote is better but i would make a little modification:

```
    Rotation and scaling are performed in the polar system by
    adding `phase_zero` (:math:`\phi_{0}`) to, and
    multiplying `modulation_zero` (:math:`M_{0}`) with
    the phase (:math:`\phi`) and modulation (:math:`M`) respectively,
    calculated from `real` and `imag`:

    .. math::
        \phi' &= \phi + \phi_{0}

        M' &= M \cdot M_{0}
```

I believe this way is easier to understand which one is added and which one is multiplied. What do you think?

cgohlke:
> I am getting some errors in testing, I believe is with the fetch datasets, but I am not sure.

You can follow the "Details" links next to the failed test runs and see exactly what failed. There's also an option to re-run failed checks.

cgohlke (approved):

cgohlke:
I just noticed that the [`skip_axes`](https://github.com/phasorpy/phasorpy/blob/d1f7d0b491b7b646330603a921df6e002940920d/src/phasorpy/phasor.py#L1057-L1059) parameter is missing from [`phasor_calibrate`](https://github.com/phasorpy/phasorpy/blob/main/src/phasorpy/phasor.py#L392-L405). Is that intentional?

It would be nice to have a tutorial using a real world dataset that can demonstrate this and other advanced features of the new `phasor_calibrate` function. I'm not aware of any suitable dataset though.

bruno-pannunzio:
I missed the `skip_axes` parameter, you are right. I will fix this in a PR. 

It would be great to have an interesting dataset that we can use for this. We could generate an suitable dataset here but we can only get .fbd files, which are not so good for a tutorial I think. But we can have something and look for files from another lab with access to other hardware.

cgohlke:
I think a small, multi-harmonic FDB dataset might be OK for this. The `skip_axes` parameter could be used on the harmonics, no? Maybe I am missing something. Anyway, it is not urgent, just something to keep track of...

PR #48, opened 2024-03-26
Add cursors module

schutyb:
## Description
Cursors module to select component in regions 
1- For circular selection 
2- For to selec component in linear ranges
...

## Release note

Summarize the changes in the code block below to be included in the
[release notes](https://www.phasorpy.org/stable/release.html):

```release-note
...
```

## Checklist

- [ ] The pull request title, summary, and description are concise.
- [ ] Related issues are linked in the description.
- [ ] New dependencies are explained.
- [ ] The source code and documentation can be distributed under the [MIT license](https://www.phasorpy.org/stable/license.html).
- [ ] The source code adheres to [code standards](https://www.phasorpy.org/stable/contributing.html#code-standards).
- [ ] New classes, functions, and features are thoroughly [tested](https://www.phasorpy.org/stable/contributing.html#tests).
- [ ] New, user-facing classes, functions, and features are [documented](https://www.phasorpy.org/stable/contributing.html#documentation).
- [ ] New features are covered in tutorials.
- [ ] No files other than source code, documentation, and project settings are added to the repository.

PR #78, opened 2024-07-01
Add function to project multi-harmonic phasor coordinates onto principal plane

cgohlke:
## Description

This PR adds a single function, `phasor_to_principal_plane`, which projects multi-harmonic phasor coordinates onto the principal plane, along which coordinate axes the phasor coordinates have the largest variations. The function is implemented using PCA, finding the two most significant eigenvectors in the covariance matrix.

The function also tries to rotate and scale the projected coordinates such that their center lies in same quadrant and the projection of [1, 0] lies at [1, 0]. This allows for a more familiar presentation. 

The transformation (projection, rotation, scaling) is affine. It preserves collinearity and ratios of distances, such that it should be possible to use the projected coordinates instead of phasor coordinates in some of PhasorPy's visualization and analysis functions.

This PR is work in progress. I am not confident that the single synthetic test is adequate. The affine transformation should also preserve convexity, but during testing I came across cases where that was not the case. Maybe there is still something wrong with the implementation or the method is numerically unstable...

See also [doi:10.1021/acs.jpclett.0c02319](https://doi.org/10.1021/acs.jpclett.0c02319). In the supplement it raises an interesting concern that (if I understood correctly) in FLIM data there is correlation in the frequency-dimension due to the presence of shot noise, which makes the projection more noisy.

## Release note

```release-note
Add function to project multi-harmonic phasor coordinates onto principal plane
```

## Checklist


cgohlke:
Apparently the synthetic test data lead to significant different results across platforms.

cgohlke:
How about renaming the function to `phasor_principal_components` and add an `ndims=2` parameter? It would allow, for example, to project the phasor coordinates onto 3 dimensions (are there any use cases?). However, it would make typing the returns, documentation, and the implementation of `reorient` more complex...

cgohlke:
I think this is ready for review.

bruno-pannunzio:
> I think this is ready for review.

Excelent @cgohlke! I will have a look at this and #81 as soon as possible, as I need to finish a couple of things first of #48 so we can move forward with the PR.

src/phasorpy/phasor.py
```diff
@@ -1934,6 +1936,150 @@ def phasor_from_fret_acceptor(
...
+        where the first axis is the frequency::
+
+            x, y = numpy.dot(
+                numpy.vstack(
+                    real.reshape(real.shape[0], -1),
+                    imag.reshape(imag.shape[0], -1),
+                ),
+                transformation_matrix,
+            ).reshape(2, *real.shape[1:])
+
```
bruno-pannunzio, line 1994: Maybe add reference to the tutorial?

Also a reference to the paper you cited in the PR description can also be helpful  to understand it's applications.

tutorials/phasorpy_principle_components.py
```diff
@@ -0,0 +1,147 @@
+"""
+Principle components
```
bruno-pannunzio, line 2: I am not sure but maybe here is Principal instead of Principle?

bruno-pannunzio (approved):
@cgohlke I added some minor comments but I think this is ready to be merged. We should test with real data as you commented in the tutorial. Maybe with hyperspectral data, as FLIM data seems to be not good due to noise (I also understood the same from the paper).

cgohlke:
Hi @bruno-pannunzio : Thanks for the review. I added the reference and corrected the spelling. Let's try to find a real dataset that could profit from this projection and apply the cursor visualization.

Do you think there could be any confusion from using the terms "components" (vs lifetime components) and "principal plane" (used in optics)? 

> How about renaming the function to `phasor_principal_components` and add an `ndims=2` parameter? It would allow, for example, to project the phasor coordinates onto 3 dimensions (are there any use cases?). However, it would make typing the returns, documentation, and the implementation of reorient more complex...

Any thought about this?

bruno-pannunzio:
Oh sorry I didn't see the previous comment. I hadn't realized but you may be right that there could be some confusion with the lifetime components. That's why I think that although `phasor_principal_components` is more accurate, the current name may lead to less confusion. Maybe the tutorial title can be Principal Component Analysis, since PCA is a commonly known method and by adding the 'Analysis' it makes it much more clear that this is PCA and not the lifetime components analysis, or something like Projection to principal plane. Maybe @lmalacrida can give is some insight into this or we can discuss next meeting. 

On the other hand, I don't think it will lead to confusion with the optical plane, since this is clearly an analysis and I am not aware of this kind of analysis related to the optical plane.

Related to the `ndims`, it would be nice maybe in the future to add the possibility to add more principal components, as it sometimes helps to include more dimensions (if they help to explain the variance of the data), but I am not aware if this is the case in the phasor analysis. I agree that it will add complexity to implementation and documentation and maybe other issues are more important at this stage.

cgohlke:
Sounds good. Let's leave this PR open for now and briefly discuss tomorrow.

cgohlke:
OK, I am going to merge this even though it is an experimental/unproven feature.

PR #84, opened 2024-07-25
Add elliptic cursors

cgohlke:
## Description

This PR adds support for elliptic cursors to the `plot.PhasorPlot.cursor` method and the `cursors` module. Tests and tutorials are updated.

Items to discuss:

- [ ] The new `mask_from_elliptic_cursor` function behaves exactly like `mask_from_circular_cursor` by default. Removing `mask_from_circular_cursor` would simplify implementation and testing.
- [x] The `radius_b` parameter of the `mask_from_elliptic_cursor` function sets the ellipse radius along the minor axis. Is the name `radius_b` appropriate?
- [x] The default `angle` of the ellipse in the `plot.PhasorPlot.cursor` method aligns with a circle around the origin or the universal semicircle depending on the `allquadrants` mode of the plot. There is no option in the `mask_from_elliptic_cursor` function to align with the universal semicircle. Is it worth adding? The `angle` can always be set explicitly.
- [ ] Phase and modulation lines are not drawn inside ellipses in the `plot.PhasorPlot.cursor` method. This would require implementing `_intersection_ellipse_circle` and  `_intersection_ellipse_line` ufuncs. Is it worth implementing this functionality?
- [ ] I am really not sure that handling multiple cursors in the `mask_from_*_cursor` is the right approach. It makes implementation, documentation, and testing more complicated than necessary, and the user has to keep track of the dimensions returned. Is there a significant use case where this simplifies things for users as compared to calling the functions in a loop or list comprehension?

## Release note

```release-note
Add elliptic cursors
```

## Checklist

### DIFF
--- a/tutorials/phasorpy_cursors.py
+++ b/tutorials/phasorpy_cursors.py
@@ -75,7 +75,7 @@ plt.show()
 # Elliptic cursors
 # ----------------
 #
-# Use elliptic cursors to mask morde defined regions of interest in the
+# Use elliptic cursors to mask more defined regions of interest in the
 # phasor space:
 
 radius = [0.1, 0.06]
--- a/tutorials/phasorpy_cursors.py
+++ b/tutorials/phasorpy_cursors.py
@@ -14,6 +14,7 @@ import matplotlib.pyplot as plt
 from phasorpy.color import CATEGORICAL
 from phasorpy.cursors import (
     mask_from_circular_cursor,
+    mask_from_elliptic_cursor,
     mask_from_polar_cursor,
     pseudo_color,
 )
@@ -28,35 +29,89 @@ from phasorpy.plot import PhasorPlot
 signal = read_lsm(fetch('paramecium.lsm'))
 mean, real, imag = phasor_from_signal(signal, axis=0)
 
+threshold = mean > 1
+
 # %%
 # Circular cursors
 # ----------------
 #
 # Use circular cursors to mask regions of interest in the phasor space:
 
-cursors_real = [-0.33, 0.55]
-cursors_imag = [-0.72, -0.72]
+cursors_real = [-0.33, 0.54]
+cursors_imag = [-0.72, -0.74]
+radius = [0.2, 0.22]
 
 circular_mask = mask_from_circular_cursor(
-    real, imag, cursors_real, cursors_imag, radius=0.2
+    real, imag, cursors_real, cursors_imag, radius=radius
 )
 
 # %%
 # Show the circular cursors in a phasor plot:
 
-threshold = mean > 1
-
 plot = PhasorPlot(allquadrants=True, title='Circular cursors')
-plot.hist2d(real[threshold], imag[threshold])
+plot.hist2d(real[threshold], imag[threshold], cmap='Greys')
 for i in range(len(cursors_real)):
     plot.cursor(
         cursors_real[i],
         cursors_imag[i],
-        radius=0.2,
+        radius=radius[i],
         color=CATEGORICAL[i],
         linestyle='-',
     )
 plot.show()
+
+# %%
+#
+# The cursor masks can be blended to produce a pseudo-colored image:
+
+pseudo_color_image = pseudo_color(*circular_mask)
+
+fig, ax = plt.subplots()
+ax.set_title('Pseudo-color image from circular cursors')
+ax.imshow(pseudo_color_image)
+plt.show()
+
+# %%
+# Elliptic cursors
+# ----------------
+#
+# Use elliptic cursors to mask morde defined regions of interest in the
+# phasor space:
+
+radius = [0.1, 0.06]
+radius_b = [0.3, 0.25]
+
+elliptic_mask = mask_from_elliptic_cursor(
+    real, imag, cursors_real, cursors_imag, radius=radius, radius_b=radius_b
+)
+
+# %%
+# Show the elliptic cursors in a phasor plot:
+
+plot = PhasorPlot(allquadrants=True, title='Elliptic cursors')
+plot.hist2d(real[threshold], imag[threshold], cmap='Greys')
+for i in range(len(cursors_real)):
+    plot.cursor(
+        cursors_real[i],
+        cursors_imag[i],
+        radius=radius[i],
+        radius_b=radius_b[i],
+        color=CATEGORICAL[i],
+        linestyle='-',
+    )
+plot.show()
+
+# %%
+#
+# The mean intensity image can be used as a base layer to overlay
+# the masks from the elliptic cursors:
+
+pseudo_color_image = pseudo_color(*elliptic_mask, intensity=mean)
+
+fig, ax = plt.subplots()
+ax.set_title('Pseudo-color image from elliptic cursors and intensity')
+ax.imshow(pseudo_color_image)
+plt.show()
 
 # %%
 # Polar cursors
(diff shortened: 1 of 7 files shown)


tutorials/phasorpy_cursors.py
```diff
@@ -28,36 +29,90 @@
...
+fig, ax = plt.subplots()
+ax.set_title('Pseudo-color image from circular cursors')
+ax.imshow(pseudo_color_image)
+plt.show()
+
+# %%
+# Elliptic cursors
+# ----------------
+#
+# Use elliptic cursors to mask morde defined regions of interest in the
```
bruno-pannunzio, line 78: I believe here is more instead of morde?

bruno-pannunzio:
Hi @cgohlke, I think you raised some interesting points for discussion. I think the input from @schutyb and @lmalacrida on this points will be helpful. 

> * The new `mask_from_elliptic_cursor` function behaves exactly like `mask_from_circular_cursor` by default. Removing `mask_from_circular_cursor` would simplify implementation and testing.

This is a good idea I think. Maybe add to the function documentation that in case the second radius parameter is `None`, then that will be treated as a circular cursor.

> * The `radius_b` parameter of the `mask_from_elliptic_cursor` function sets the ellipse radius along the minor axis. Is the name `radius_b` appropriate?

I think a better name would be `major_radius` and `minor_radius`. That is much more explicit or their meaning. Although I know that they can be interchanged in this case, so I am not exactly sure if the names make sense. Also in the case of the circular cursor it doesn't make a lot of sense. I don't know how other libraries that use ellipses name their radius parameters, but maybe we can think of something similar.

> * The default `angle` of the ellipse in the `plot.PhasorPlot.cursor` method aligns with a circle around the origin or the universal semicircle depending on the `allquadrants` mode of the plot. There is no option in the `mask_from_elliptic_cursor` function to align with the universal semicircle. Is it worth adding? The `angle` can always be set explicitly.

I think it will help for consistency, but it will depend on how much time the implementation will take as I don't think is something too necessary.

> * Phase and modulation lines are not drawn inside ellipses in the `plot.PhasorPlot.cursor` method. This would require implementing `_intersection_ellipse_circle` and  `_intersection_ellipse_line` ufuncs. Is it worth implementing this functionality?

I don't think this is necessary.

> * I am really not sure that handling multiple cursors in the `mask_from_*_cursor` is the right approach. It makes implementation, documentation, and testing more complicated than necessary, and the user has to keep track of the dimensions returned. Is there a significant use case where this simplifies things for users as compared to calling the functions in a loop or list comprehension?

I think my arguments were in the previous PR, but if you believe this will help with simplicity, we can remove the handling of multiple cursors at one. Eventually, it isn't that much of a difference as you said with using loops or list comprehension. It will be:

`polar_masks = [
    mask_from_polar_cursor(real, imag, phase_min[i], phase_max[i], modulation_min[i], modulation_max[i])
    for i in range(len(phase_min))
]`

instead of:

`polar_mask = mask_from_polar_cursor(
    real, imag, phase_min, phase_max, modulation_min, modulation_max
)`

bruno-pannunzio:
> I think my arguments were in the previous PR, but if you believe this will help with simplicity, we can remove the handling of multiple cursors at one. Eventually, it isn't that much of a difference as you said with using loops or list comprehension. It will be:
> 
> `polar_mask = [ mask_from_polar_cursor(real, imag, phase_min[i], phase_max[i], modulation_min[i], modulation_max[i]) for i in range(len(phase_min)) ]`
> 
> instead of:
> 
> `polar_mask = mask_from_polar_cursor(real, imag, phase_min, phase_max, modulation_min, modulation_max)`

Also, I forgot to mention that I don't know if this will have an impact in performance with large images, but is something we might be able to test if it's worth it or not.

cgohlke:
> polar_mask = [ mask_from_polar_cursor(real, imag, phase_min[i], phase_max[i], modulation_min[i], modulation_max[i]) for i in range(len(phase_min)) ]
>
> instead of:
>
> polar_mask = mask_from_polar_cursor(real, imag, phase_min, phase_max, modulation_min, modulation_max)


> Also, I forgot to mention that I don't know if this will have an impact in performance with large images, but is something we might be able to test if it's worth it or not.

One does not have to use loops or comprehension as the "single" cursor functions would allow for broadcasting. The question is if it is worth sparing the user from making the input broadcastable for a common case and thereby limiting the function for more advanced (rare?) broadcasting cases. Let's leave this as is for now and see how well it works for real usage.

cgohlke:
>  I don't know how other libraries that use ellipses name their radius parameters

The obvious choice for an ellipse would be `width` and `height`, but that does not align with the `mask_from_circular_cursor ` function and the `PhasorPlot.cursor` method. A while back we agreed on using radius instead of diameter in the library. Should we review that decision?

cgohlke:
- renamed `radius_b` to `radius_minor`.

- added an `align_semicircle` parameter to the functions, which allows switching from the default "align to unit circle" mode to "align to semicircle". The parameter is not vectorized, applies to all ellipses.

I don't think the other discussion items should wait for feedback from others. This PR is required to move on to the interactive cursor app.

schutyb:
> * [ ]  The new `mask_from_elliptic_cursor` function behaves exactly like `mask_from_circular_cursor` by default. Removing `mask_from_circular_cursor` would simplify implementation and testing.
Agreed


> * [ ]  Phase and modulation lines are not drawn inside ellipses in the `plot.PhasorPlot.cursor` method. This would require implementing `_intersection_ellipse_circle` and  `_intersection_ellipse_line` ufuncs. Is it worth implementing this functionality?
Personally I don´t think is necesary, I´m using phasorpy.plot in the paper we are writing now with @lmalacrida those lines from the circular cursors make a bit confusing the image, mainly when you have many cursors.

schutyb:
> * [ ]  The new `mask_from_elliptic_cursor` function behaves exactly like `mask_from_circular_cursor` by default. Removing `mask_from_circular_cursor` would simplify implementation and testing.

Totally agreed. 

> * [ ]  Phase and modulation lines are not drawn inside ellipses in the `plot.PhasorPlot.cursor` method. This would require implementing `_intersection_ellipse_circle` and  `_intersection_ellipse_line` ufuncs. Is it worth implementing this functionality?

Personally I don´t think is necesary, I´m using phasorpy.plot in the paper we are writing now with @lmalacrida those lines from the circular cursors make a bit confusing the image, mainly when you have many cursors.

schutyb:
> > I don't know how other libraries that use ellipses name their radius parameters
> 
> The obvious choice for an ellipse would be `width` and `height`, but that does not align with the `mask_from_circular_cursor ` function and the `PhasorPlot.cursor` method. A while back we agreed on using radius instead of diameter in the library. Should we review that decision?

I think this is more appropiate when `width` and `height` are equal it´s a circle. At the begining we started we radius bc it the parameter for the circle, but a circle is an ellipse.

cgohlke:
> I´m using phasorpy.plot in the paper we are writing now with @lmalacrida those lines from the circular cursors make a bit confusing the image, mainly when you have many cursors.

Then draw circles. There's a method for that in the PhasorPlot class. The cursor method is to draw graphical cursors in phasor space: a polar crosshair by default, which can optionally be extended to a range (polar rectangle), or limited by circular or now elliptic shapes.

bruno-pannunzio (approved):
I agree @cgohlke with your comments and the changes you made. I think we all agreed that removing the `mask_from_circular_cursor` is the best approach to simplify the module, so maybe we can do that in another PR if you believe this is ready and wan't to merge it. I think we can merge and then make the corresponding changes later so you can continue with the interactive cursor app.

schutyb (approved):

PR #63, opened 2024-04-29
Add LFD workshop FLIM tutorial

cgohlke:
## Description

This tutorial is a close adaptation of the [LFD Workshop computer training tutorial - FLIM section](https://www.lfd.uci.edu/workshop/2021/files/LFDWorkshop2021-ComputerTraining.pdf) (by E. Gratton, M. Digman, and J. Unruh) using the PhasorPy library instead of the [Globals for Images · SimFCS](https://www.lfd.uci.edu/globals/) and Excel software.

I was curious how much of the original tutorial could be implemented using the current state of the PhasorPy library. 

The following functionality is still missing from PhasorPy to finish the tutorial:

- [ ] synthesize time-domain signal (#6, #86)
- [ ] filtering, thresholding images (#85, #88)
- [ ] cursors (#48)
- [ ] component analysis (#47 and #59)

## Release note

Summarize the changes in the code block below to be included in the
[release notes](https://www.phasorpy.org/stable/release.html):

```release-note
Add LFD workshop FLIM tutorial
```

## Checklist

- [x] The pull request title, summary, and description are concise.
- [x] Related issues are linked in the description.
- [ ] New dependencies are explained.
- [ ] The source code and documentation can be distributed under the [MIT license](https://www.phasorpy.org/stable/license.html).
- [x] The source code adheres to [code standards](https://www.phasorpy.org/stable/contributing.html#code-standards).
- [ ] New classes, functions, and features are thoroughly [tested](https://www.phasorpy.org/stable/contributing.html#tests).
- [ ] New, user-facing classes, functions, and features are [documented](https://www.phasorpy.org/stable/contributing.html#documentation).
- [ ] New features are covered in tutorials.
- [x] No files other than source code, documentation, and project settings are added to the repository.

PR #88, opened 2024-08-01
Add `phasor_threshold` function to `phasor` module.

bruno-pannunzio:
## Description
This PR adds the function `phasor_threshold` to `phasor` module to apply a mask and replace sub-threshold `mean` values  in `mean`, `real`, and `imag`.

This functionality is proposed in #21. 

## Release note

```release-note
Add `phasor_threshold` function.
```

## Checklist

### DIFF
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -135,14 +135,14 @@ def _phasor_from_signal(
                             dc = dc + sample
                             re = re + sample * sincos[h, k, 0]
                             im = im + sample * sincos[h, k, 1]
-                        if dc > 1e-16:
+                        if dc != 0.0:
                             re = re / dc
                             im = im / dc
                             dc = dc /samples
                         else:
                             dc = 0.0
-                            re = 0.0  # inf?
-                            im = 0.0  # inf?
+                            re = NAN if re == 0.0 else re * INFINITY
+                            im = NAN if im == 0.0 else im * INFINITY
                         if h == 0:
                             mean[i, j] = <float_t> dc
                         real[h, i, j] = <float_t> re
@@ -163,14 +163,14 @@ def _phasor_from_signal(
                             dc = dc + sample
                             re = re + sample * sincos[h, k, 0]
                             im = im + sample * sincos[h, k, 1]
-                        if dc > 1e-16:
+                        if dc != 0.0:
                             re = re / dc
                             im = im / dc
                             dc = dc /samples
                         else:
                             dc = 0.0
-                            re = 0.0  # inf?
-                            im = 0.0  # inf?
+                            re = NAN if re == 0.0 else re * INFINITY
+                            im = NAN if im == 0.0 else im * INFINITY
                         if h == 0:
                             mean[i, j] = <float_t> dc
                         real[h, i, j] = <float_t> re
@@ -191,14 +191,14 @@ def _phasor_from_signal(
                             dc += sample
                             re += sample * sincos[h, k, 0]
                             im += sample * sincos[h, k, 1]
-                        if dc > 1e-16:
+                        if dc != 0.0:
                             re /= dc
                             im /= dc
                             dc /= samples
                         else:
                             dc = 0.0
-                            re = 0.0  # inf?
-                            im = 0.0  # inf?
+                            re = NAN if re == 0.0 else re * INFINITY
+                            im = NAN if im == 0.0 else im * INFINITY
                         if h == 0:
                             mean[i, j] = <float_t> dc
                         real[h, i, j] = <float_t> re
@@ -282,7 +282,7 @@ def _phasor_from_lifetime(
                     else:
                         for s in range(ncomp):
                             sum += fraction[t, s]
-                    if fabs(sum) < 1e-16:
+                    if fabs(sum) < 1e-15:
                         phasor[0, f, t] = <float_t> NAN
                         phasor[1, f, t] = <float_t> NAN
                         continue
@@ -315,7 +315,7 @@ def _phasor_from_lifetime(
                         sum = 0.0
                         for s in range(ncomp):
                             sum += fraction[0, s] * lifetime[t, s]  # Fdc
-                    if fabs(sum) < 1e-16:
+                    if fabs(sum) < 1e-15:
                         phasor[0, f, t] = <float_t> NAN
                         phasor[1, f, t] = <float_t> NAN
                         continue
@@ -351,7 +351,7 @@ def _phasor_from_lifetime(
                     else:
                         for s in range(ncomp):
                             sum += fraction[t, s]
-                    if fabs(sum) < 1e-16:
+                    if fabs(sum) < 1e-15:
                         phasor[0, f, t] = <float_t> NAN
                         phasor[1, f, t] = <float_t> NAN
                         continue
@@ -594,7 +594,7 @@ cdef inline (double, double) linear_combination(
     int1 *= frac
     int2 *= 1.0 - frac
     frac = int1 + int2
-    if frac == 0.0:
+    if fabs(frac) < 1e-15:
         return real, imag
     return (
         (int1 * real1 + int2 * real2) / frac,
@@ -795,7 +795,11 @@ cdef (float_t, float_t) _polar_from_reference(
 ) noexcept nogil:
     """Return polar coordinates for calibration from reference coordinates."""
     if fabs(measured_modulation) == 0.0:
-        return known_phase - measured_phase, <float_t> INFINITY
+        # return known_phase - measured_phase, <float_t> INFINITY
+        return (
+            known_phase - measured_phase,
+            <float_t> (NAN if known_modulation == 0.0 else INFINITY)
+        )
     return known_phase - measured_phase, known_modulation / measured_modulation
 
 
@@ -814,7 +818,11 @@ cdef (float_t, float_t) _polar_from_reference_phasor(
         double known_modulation = hypot(known_real, known_imag)
 
     if fabs(measured_modulation) == 0.0:
-        return <float_t> (known_phase - measured_phase), <float_t> INFINITY
+        # return <float_t> (known_phase - measured_phase), <float_t> INFINITY
+        return (
+            <float_t> (known_phase - measured_phase),
+            <float_t> (NAN if known_modulation == 0.0 else INFINITY)
+        )
     return (
         <float_t> (known_phase - measured_phase),
         <float_t> (known_modulation / measured_modulation)
@@ -1429,9 +1437,9 @@ cdef (double, double, double, double) _intersection_circle_line(
     rdd = sqrt(rdd)
     return (
         x + (dd * dy + copysign(1.0, dy) * dx * rdd) / dr,
-        y + (-dd * dx + abs(dy) * rdd) / dr,
+        y + (-dd * dx + fabs(dy) * rdd) / dr,
         x + (dd * dy - copysign(1.0, dy) * dx * rdd) / dr,
-        y + (-dd * dx - abs(dy) * rdd) / dr,
+        y + (-dd * dx - fabs(dy) * rdd) / dr,
     )
 
 ###############################################################################
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -1506,28 +1506,67 @@ cdef float_t _blend_lighten(
     return <float_t> max(a, b)
 
 ###############################################################################
-# Threshold ufunc
+# Threshold ufuncs
 
 
 @cython.ufunc
-cdef (double, double, double) _mask_threshold(
+cdef (double, double, double) _phasor_threshold(
     float_t mean,
     float_t real,
     float_t imag,
-    float_t mean_lower,
-    float_t mean_upper,
-    float_t phasor_lower,
-    float_t phasor_upper,
+    float_t mean_min,
+    float_t mean_max,
+    float_t real_min,
+    float_t real_max,
+    float_t imag_min,
+    float_t imag_max,
     float_t fill,
 ) noexcept nogil:
     """Return thresholded values."""
-    cdef double nan = NAN
-
     if isnan(mean) or isnan(real) or isnan(imag):
-        return nan, nan, nan
-    if (mean_lower < mean < mean_upper and
-            phasor_lower < real < phasor_upper and
-            phasor_lower < imag < phasor_upper):
+        return NAN, NAN, NAN
+    if (
+        mean_min <= mean < mean_max
+        and real_min <= real < real_max
+        and imag_min <= imag < imag_max
+    ):
         return mean, real, imag
-    else:
-        return fill, fill, fill
+    return fill, fill, fill
+
+
+@cython.ufunc
+cdef (double, double, double) _phasor_polar_threshold(
+    float_t mean,
+    float_t real,
+    float_t imag,
+    float_t mean_min,
+    float_t mean_max,
+    float_t real_min,
+    float_t real_max,
+    float_t imag_min,
+    float_t imag_max,
+    float_t phase_min,
+    float_t phase_max,
+    float_t modulation_min,
+    float_t modulation_max,
+    float_t fill,
+) noexcept nogil:
+    """Return thresholded values."""
+    cdef:
+        float_t phi
+        float_t mod
+    if isnan(mean) or isnan(real) or isnan(imag):
+        return NAN, NAN, NAN
+    if (
+        mean_min <= mean < mean_max
+        and real_min <= real < real_max
+        and imag_min <= imag < imag_max
+    ):
+        phi = atan2(imag, real)
+        mod = sqrt(real * real + imag * imag)
+        if (
+            phase_min <= phi < phase_max
+            and modulation_min <= mod < modulation_max
+        ):
+            return mean, real, imag
+    return fill, fill, fill
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -7,16 +7,11 @@
 
 """Cython implementation of low-level functions for the PhasorPy package."""
 
-# these lines must be at top of module
-# TODO: https://github.com/cython/cython/issues/6064
-
 # TODO: replace short with unsigned char when Cython supports it
 # https://github.com/cython/cython/pull/6196#issuecomment-2209509572
 
-cimport numpy
-
-numpy.import_array()
-numpy.import_ufunc()
+# TODO: use fused return types for functions returning more than two items
+# https://github.com/cython/cython/issues/6328
 
 cimport cython
 
@@ -30,7 +25,9 @@ from libc.math cimport (
     atan2,
     copysign,
     cos,
+    exp,
     fabs,
+    floor,
     hypot,
     isnan,
     sin,
@@ -209,7 +206,7 @@ def _phasor_from_signal(
 
 
 def _phasor_from_lifetime(
-    double[:, :, ::1] phasor,
+    float_t[:, :, ::1] phasor,
     const double[::1] frequency,
     const double[:, ::1] lifetime,
     const double[:, ::1] fraction,
@@ -250,7 +247,6 @@ def _phasor_from_lifetime(
         ssize_t ntau = lifetime.shape[0]  # number lifetimes
         ssize_t nfrac = fraction.shape[0]  # number fractions
         double twopi = 2.0 * M_PI * unit_conversion
-        double nan = NAN  # float('NaN')
         double freq, tau, frac, sum, re, im, gs
         ssize_t f, t, s
 
@@ -265,8 +261,8 @@ def _phasor_from_lifetime(
         # scalar
         tau = lifetime[0, 0] * frequency[0] * twopi  # omega_tau
         gs = 1.0 / (1.0 + tau * tau)
-        phasor[0, 0, 0] = gs
-        phasor[1, 0, 0] = gs * tau
+        phasor[0, 0, 0] = <float_t> gs
+        phasor[1, 0, 0] = <float_t> (gs * tau)
         return
 
     if ntau == nfrac:
@@ -287,8 +283,8 @@ def _phasor_from_lifetime(
                         for s in range(ncomp):
                             sum += fraction[t, s]
                     if fabs(sum) < 1e-16:
-                        phasor[0, f, t] = nan
-                        phasor[1, f, t] = nan
+                        phasor[0, f, t] = <float_t> NAN
+                        phasor[1, f, t] = <float_t> NAN
                         continue
                     for s in range(ncomp):
                         tau = lifetime[t, s]
@@ -299,8 +295,8 @@ def _phasor_from_lifetime(
                         gs = frac / (1.0 + tau * tau)
                         re += gs
                         im += gs * tau
-                    phasor[0, f, t] = re
-                    phasor[1, f, t] = im
+                    phasor[0, f, t] = <float_t> re
+                    phasor[1, f, t] = <float_t> im
         return
 
     if ntau > 1 and nfrac == 1:
@@ -320,8 +316,8 @@ def _phasor_from_lifetime(
                         for s in range(ncomp):
                             sum += fraction[0, s] * lifetime[t, s]  # Fdc
                     if fabs(sum) < 1e-16:
-                        phasor[0, f, t] = nan
-                        phasor[1, f, t] = nan
+                        phasor[0, f, t] = <float_t> NAN
+                        phasor[1, f, t] = <float_t> NAN
                         continue
                     re = 0.0
                     im = 0.0
@@ -334,8 +330,8 @@ def _phasor_from_lifetime(
                         gs = frac / (1.0 + tau * tau)
                         re += gs
                         im += gs * tau
-                    phasor[0, f, t] = re
-                    phasor[1, f, t] = im
+                    phasor[0, f, t] = <float_t> re
+                    phasor[1, f, t] = <float_t> im
         return
 
     if ntau == 1 and nfrac > 1:
@@ -356,8 +352,8 @@ def _phasor_from_lifetime(
                         for s in range(ncomp):
                             sum += fraction[t, s]
                     if fabs(sum) < 1e-16:
-                        phasor[0, f, t] = nan
-                        phasor[1, f, t] = nan
+                        phasor[0, f, t] = <float_t> NAN
+                        phasor[1, f, t] = <float_t> NAN
                         continue
                     for s in range(ncomp):
                         tau = lifetime[0, s]
@@ -368,13 +364,40 @@ def _phasor_from_lifetime(
                         gs = frac / (1.0 + tau * tau)
                         re += gs
                         im += gs * tau
-                    phasor[0, f, t] = re
-                    phasor[1, f, t] = im
+                    phasor[0, f, t] = <float_t> re
+                    phasor[1, f, t] = <float_t> im
         return
 
     raise ValueError(
         f'{lifetime.shape[0]=} and {fraction.shape[0]=} do not match'
     )
+
+
+def _gaussian_signal(
+    float_t[::1] signal,
+    const double mean,
+    const double stdev,
+):
+    """Return normal distribution, wrapped around at borders."""
+    cdef:
+        ssize_t samples = signal.shape[0]
+        ssize_t folds = 1  # TODO: calculate from stddev and samples
+        ssize_t i
+        double t, c
+
+    if stdev <= 0.0 or samples < 1:
+        return
+
+    with nogil:
+        c = 1.0 / sqrt(2.0 * M_PI) * stdev
+
+        for i in range(-folds * samples, (folds + 1) * samples):
+            t = (<double> i - mean) / stdev
+            t *= t
+            t = c * exp(-t / 2.0)
+            # i %= samples
+            i -= samples * <ssize_t> floor(<double> i / samples)
+            signal[i] += <float_t> t
 
 
 ###############################################################################
@@ -578,11 +601,10 @@ cdef inline (double, double) linear_combination(
         (int1 * imag1 + int2 * imag2) / frac
     )
 
-
 ###############################################################################
 # Phasor conversions
 
-cdef inline (double, double) phasor_from_lifetime(
+cdef inline (float_t, float_t) phasor_from_lifetime(
     float_t lifetime,
     float_t omega,
 ) noexcept nogil:
@@ -592,10 +614,10 @@ cdef inline (double, double) phasor_from_lifetime(
         double mod = 1.0 / sqrt(1.0 + t * t)
         double phi = atan(t)
 
-    return mod * cos(phi), mod * sin(phi)
+    return <float_t> (mod * cos(phi)), <float_t> (mod * sin(phi))
 
 
-cdef inline (double, double) phasor_to_apparent_lifetime(
+cdef inline (float_t, float_t) phasor_to_apparent_lifetime(
     float_t real,
     float_t imag,
     float_t omega,
@@ -614,10 +636,10 @@ cdef inline (double, double) phasor_to_apparent_lifetime(
         else:
             taumod = 0.0
 
-    return tauphi, taumod
+    return <float_t> tauphi, <float_t> taumod
 
 
-cdef inline (double, double) phasor_from_apparent_lifetime(
+cdef inline (float_t, float_t) phasor_from_apparent_lifetime(
     float_t tauphi,
     float_t taumod,
     float_t omega,
@@ -628,11 +650,11 @@ cdef inline (double, double) phasor_from_apparent_lifetime(
         double mod = 1.0 / sqrt(1.0 + t * t)
         double phi = atan(omega * tauphi)
 
-    return mod * cos(phi), mod * sin(phi)
+    return <float_t> (mod * cos(phi)), <float_t> (mod * sin(phi))
 
 
 @cython.ufunc
-cdef (double, double) _phasor_transform(
+cdef (float_t, float_t) _phasor_transform(
     float_t real,
     float_t imag,
     float_t angle,
@@ -643,11 +665,11 @@ cdef (double, double) _phasor_transform(
         double g = scale * cos(angle)
         double s = scale * sin(angle)
 
-    return real * g - imag * s, real * s + imag * g
+    return <float_t> (real * g - imag * s), <float_t> (real * s + imag * g)
 
 
 @cython.ufunc
-cdef (double, double) _phasor_transform_const(
+cdef (float_t, float_t) _phasor_transform_const(
     float_t real,
     float_t imag,
     float_t real2,
@@ -658,25 +680,31 @@ cdef (double, double) _phasor_transform_const(
 
 
 @cython.ufunc
-cdef (double, double) _phasor_to_polar(
+cdef (float_t, float_t) _phasor_to_polar(
     float_t real,
     float_t imag,
 ) noexcept nogil:
     """Return polar from phasor coordinates."""
-    return atan2(imag, real), sqrt(real * real + imag * imag)
+    return (
+        <float_t> atan2(imag, real),
+        <float_t> sqrt(real * real + imag * imag)
+    )
 
 
 @cython.ufunc
-cdef (double, double) _phasor_from_polar(
+cdef (float_t, float_t) _phasor_from_polar(
     float_t phase,
     float_t modulation,
 ) noexcept nogil:
     """Return phasor from polar coordinates."""
-    return modulation * cos(phase), modulation * sin(phase)
+    return (
+        modulation * <float_t> cos(phase),
+        modulation * <float_t> sin(phase)
+    )
 
 
 @cython.ufunc
-cdef (double, double) _phasor_to_apparent_lifetime(
+cdef (float_t, float_t) _phasor_to_apparent_lifetime(
     float_t real,
     float_t imag,
     float_t omega,
@@ -686,7 +714,7 @@ cdef (double, double) _phasor_to_apparent_lifetime(
 
 
 @cython.ufunc
-cdef (double, double) _phasor_from_apparent_lifetime(
+cdef (float_t, float_t) _phasor_from_apparent_lifetime(
     float_t tauphi,
     float_t taumod,
     float_t omega,
@@ -696,7 +724,7 @@ cdef (double, double) _phasor_from_apparent_lifetime(
 
 
 @cython.ufunc
-cdef (double, double) _phasor_from_single_lifetime(
+cdef (float_t, float_t) _phasor_from_single_lifetime(
     float_t lifetime,
     float_t omega,
 ) noexcept nogil:
@@ -706,11 +734,11 @@ cdef (double, double) _phasor_from_single_lifetime(
         double phi = atan(t)
         double mod = 1.0 / sqrt(1.0 + t * t)
 
-    return mod * cos(phi), mod * sin(phi)
+    return <float_t> (mod * cos(phi)), <float_t> (mod * sin(phi))
 
 
 @cython.ufunc
-cdef (double, double) _polar_from_single_lifetime(
+cdef (float_t, float_t) _polar_from_single_lifetime(
     float_t lifetime,
     float_t omega,
 ) noexcept nogil:
@@ -718,11 +746,11 @@ cdef (double, double) _polar_from_single_lifetime(
     cdef:
         double t = omega * lifetime
 
-    return atan(t), 1.0 / sqrt(1.0 + t * t)
+    return <float_t> atan(t), <float_t> (1.0 / sqrt(1.0 + t * t))
 
 
 @cython.ufunc
-cdef (double, double) _polar_to_apparent_lifetime(
+cdef (float_t, float_t) _polar_to_apparent_lifetime(
     float_t phase,
     float_t modulation,
     float_t omega,
@@ -739,11 +767,11 @@ cdef (double, double) _polar_to_apparent_lifetime(
             taumod = sqrt(1.0 / t - 1.0) / omega
         else:
             taumod = 0.0
-    return tauphi, taumod
+    return <float_t> tauphi, <float_t> taumod
 
 
 @cython.ufunc
-cdef (double, double) _polar_from_apparent_lifetime(
+cdef (float_t, float_t) _polar_from_apparent_lifetime(
     float_t tauphi,
     float_t taumod,
     float_t omega,
@@ -752,11 +780,14 @@ cdef (double, double) _polar_from_apparent_lifetime(
     cdef:
         double t = omega * taumod
 
-    return atan(omega * tauphi), 1.0 / sqrt(1.0 + t * t)
+    return (
+        <float_t> (atan(omega * tauphi)),
+        <float_t> (1.0 / sqrt(1.0 + t * t))
+    )
 
 
 @cython.ufunc
-cdef (double, double) _polar_from_reference(
+cdef (float_t, float_t) _polar_from_reference(
     float_t measured_phase,
     float_t measured_modulation,
     float_t known_phase,
@@ -764,12 +795,12 @@ cdef (double, double) _polar_from_reference(
 ) noexcept nogil:
     """Return polar coordinates for calibration from reference coordinates."""
     if fabs(measured_modulation) == 0.0:
-        return known_phase - measured_phase, INFINITY
+        return known_phase - measured_phase, <float_t> INFINITY
     return known_phase - measured_phase, known_modulation / measured_modulation
 
 
 @cython.ufunc
-cdef (double, double) _polar_from_reference_phasor(
+cdef (float_t, float_t) _polar_from_reference_phasor(
     float_t measured_real,
     float_t measured_imag,
     float_t known_real,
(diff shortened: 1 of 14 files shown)


bruno-pannunzio:
Hi @cgohlke! I implemented this simple function to apply the threshold as we have been discussing. I have some issues I would like to discuss before proceeding:

- [ ] As we discussed in #85, I think replacing values sub-threshold with `NaN` in the `mean`, `real` and `imag` makes sense, but we can discuss other outcomes of thresholding.
- [ ] The name of the function. You proposed `phasor_mask` in the tutorial. Maybe your idea was to have an argument that would allow to mask the average signal and phasor coordinates with a desired value, instead of the default `NaN` that we have been discussing. In that case it makes more sense the name `phasor_mask`, but if not I think `phasor_threshold` is more directly related to the use of this function.
- [ ] I applied the mask in place. An alternative would be `numpy.where` which I believe can be faster in cases of large images, but requires more memory. What do you think is better to prioritize in this case?
- [ ] Do you think is necessary to check if `mean`, `real` and `imag` are broadcastable or should we left that error to be raised by numpy in that case?

bruno-pannunzio:
Also I forgot to mention. With this implemention, if there are `NaN` present in the `mean` array, they are ignored. So if that value is not `NaN` in the `real` or `imag` arrays, they are not replaced by `NaN`. Do you think it makes sense to force those values to be `NaN`? I am thinking if it exists a case where some pixel is `NaN` in the `mean` and not in `real` and `imag` but I am not sure it is possible.

cgohlke:
Let's keep the name `phasor_threshold` for now and discuss other items.

First, let's agree on what the threshold function should do in a common PhasorPy based workflow: to filter out values from the phasor coordinates that should be ignored in later processing and visualization. Also to be used as a one-stop function to sanitize phasor coordinates coming from different sources. No?

Besides special values/NaN we may want to consider [numpy masked array](https://numpy.org/doc/stable/reference/maskedarray.html).

I think the current threshold function is too narrow in scope. It is also common to apply upper intensity thresholds (e.g. to filter saturated pixels or INF values) and/or to apply lower and higher thresholds for phasor and/or polar coordinates. 

The threshold filter should not be applied in-place by default. Instead, add an `out` parameter that can be set to the input arrays to achieve in-place behavior. That is a common API in the numpy ecosystem.

Regarding NaNs being present in any input array: Probably any NaN in any input array should propagate to all output arrays. But I have not really thought it through.

Regarding broadcasting: the function should ideally behave like a numpy ufunc since it only operates on one array element a time.

The out parameter, NaN/INF handling, and ufunc behavior can all be implemented easily via a Cython ufunc.

bruno-pannunzio:
Thanks @cgohlke for your comments. There are many aspects I didn't think before but make total sense to consider for thresholding. 

I will look into all the aspects you raised. I think we can discuss next meeting how should we handle in general the `NaN`s and also in this case in particular.

cgohlke:
It might also be necessary to include multiple harmonics in the thresholding (for example, in the [LFD workshop tutorial](https://github.com/phasorpy/phasorpy/pull/63/files#diff-0a88cd6363d8a0d8476f23053022818d219fe71387b13b352c33b45d34d038b6R207-R215)).

bruno-pannunzio:
Hi @cgohlke. I have been researching if it's convenient to use `numpy.ma.masked_array` in this case, and maybe in other features of the library. If I understood correctly, the `masked_array` element contains data of the original array, the boolean mask and the fill value to replace the masked out elements. Many numpy functions work well with `masked_array` and ignore masked out values, for example `numpy.mean` doesn't consider masked out values if the input is a `masked_array`, which could be great in our case. 

The problem I see is that `numpy.asarray` which is extensively used in the library, converts the `masked_array` to an array without applying the mask. So this would render the mask useless in many cases. On the other hand, `numpy.atleast_1d` does preserved the `masked_array`, but we may not want to replace all `numpy.asarray` by `numpy.atleast_1d` since they do exactly the same. Also we would need to check thoroughly in all functions that the behaviour of the masked array is as expected, which at some point we will have to do if we mask with `NaN`, but anyway, it can be a lot of work. Also, an alternative would be to apply the mask with `numpy.filled`, but that would be the same as the more conventional approach and don't see much benefit. Please, when you have a moment let me know what you think about this.

I am also encountering not so straight forward the implementation of the thresholding function with cython considering all the requirements you suggested before, which I think it should have. So I will update with the current state of the implementation asking for your input since you are more familiar with cython and can be of great help.

cgohlke:
> The problem I see is that numpy.asarray which is extensively used in the library, converts the masked_array to an array without applying the mask. So this would render the mask useless in many cases. On the other hand, numpy.atleast_1d does preserved the masked_array, but we may not want to replace all numpy.asarray by numpy.atleast_1d

We could switch to [numpy.asanyarray](https://numpy.org/doc/stable/reference/generated/numpy.asanyarray.html). 

> I am also encountering not so straight forward the implementation of the thresholding function with cython considering all the requirements you suggested before.

I can see that handling multiple harmonics might be challenging. It doesn't really fit into an ufunc. Maybe it can be handled in a separate function or left out for now?

bruno-pannunzio:
@cgohlke I updated the PR with the Cython implementation, and some other modifications you suggested before.

> I can see that handling multiple harmonics might be challenging. It doesn't really fit into an ufunc. Maybe it can be handled in a separate function or left out for now?

This is actually handled with the current implementation. Numpy broadcasting is used in the case for more than one harmonic, as long as the `phasor_lower` and `phasor_upper` parameters have the same dimensions as the phasor coordinates, this can be broadcasted along the harmonic axis.

There are a couple of things I would like to get your input in particular also:

1. I wasn't able to implement correctly the `out` parameter to allow in place modification of the arrays. I am not sure how to do this with cython ufuncs. I tried and searched for solutions but I couldn't find one. Maybe you can give some insight on how I can approach this. At the moment, it is not performing in place modifications if the `out` is the same as the input arrays.
2. The lower and upper bounds for thresholding. Do you think it makes sense to have this many arguments? The lower and upper bounds for thresholding phasor coordinates is the same for `real` and `imag`, do you think this should be separate or is it a more efficent way to implement the arguments for bounds?
3. I am not entirely convinced of the naming of the arguments.

cgohlke:
> This is actually handled with the current implementation.

Nice.

> I wasn't able to implement correctly the out parameter to allow in place modification of the arrays

There should be no need to handle the `out` parameter. Just forward any `**kwargs` to the ufunc for example like in [`phasor_multiply`](https://github.com/phasorpy/phasorpy/blob/3106ce79e60a2643bd94b4834f899e291ad96e3d/src/phasorpy/phasor.py#L850-L908)

> The lower and upper bounds for thresholding. Do you think it makes sense to have this many arguments?

How about making the function as flexible as possible and have `_lower` and `_upper` parameters for `mean`, `real`, `imag`, `phase`, and `modulation`? That should be straightforward to implement in one ufunc, but if that does not work well (speed?) there could be simpler ufuncs for common cases that get dispatched in the main function. Currently `fill_value` stands out, because I don't see a need to ever varying values in one function call. Unfortunately ufuncs do not support constants AFAICT.

src/phasorpy/phasor.py
```diff
@@ -67,6 +67,10 @@
 
   - :py:func:`phasor_filter`
 
+- threshold phasor coordinates based on mean intensity:
```
cgohlke, line 70: In a way, it is a filter function. Move it in one category with `phasor_filter`?

src/phasorpy/phasor.py
```diff
@@ -2724,6 +2730,76 @@ def phasor_filter(
...
+    real: ArrayLike,
+    imag: ArrayLike,
+    /,
+    mean_lower: float = -numpy.inf,
+    mean_upper: float = numpy.inf,
+    *,
+    phasor_lower: float = -numpy.inf,
+    phasor_upper: float = numpy.inf,
+    fill_value: Any = numpy.nan,
+    out: tuple[NDArray[Any], NDArray[Any], NDArray[Any]] | None = None,
```
cgohlke, line 2744: Use `**kwargs: Any,` instead and forward that to the ufunc.
```
    **kwargs
        Optional `arguments passed to numpy universal functions
        <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
```

src/phasorpy/phasor.py
```diff
@@ -2724,6 +2730,76 @@ def phasor_filter(
...
+) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
+    """Return phasor coordinates with mean below threshold set to NaN.
```
cgohlke, line 2746: Return phasor coordinates with out of range values replaced by NaN.

src/phasorpy/_phasorpy.pyx
```diff
@@ -1504,3 +1504,30 @@ cdef float_t _blend_lighten(
...
+    float_t real,
+    float_t imag,
+    float_t mean_lower,
+    float_t mean_upper,
+    float_t phasor_lower,
+    float_t phasor_upper,
+    float_t fill,
+) noexcept nogil:
+    """Return thresholded values."""
+    cdef double nan = NAN
```
cgohlke, line 1524: No need to define this. Just `return NAN, NAN, NAN`

src/phasorpy/_phasorpy.pyx
```diff
@@ -1504,3 +1504,30 @@ cdef float_t _blend_lighten(
...
+
+    if isnan(mean) or isnan(real) or isnan(imag):
+        return nan, nan, nan
+    if (mean_lower < mean < mean_upper and
+            phasor_lower < real < phasor_upper and
+            phasor_lower < imag < phasor_upper):
+        return mean, real, imag
+    else:
```
cgohlke, line 1532: `else:` not necessary.

src/phasorpy/_phasorpy.pyx
```diff
@@ -1504,3 +1504,30 @@ cdef float_t _blend_lighten(
...
+    float_t fill,
+) noexcept nogil:
+    """Return thresholded values."""
+    cdef double nan = NAN
+
+    if isnan(mean) or isnan(real) or isnan(imag):
+        return nan, nan, nan
+    if (mean_lower < mean < mean_upper and
+            phasor_lower < real < phasor_upper and
+            phasor_lower < imag < phasor_upper):
```
cgohlke, line 1530: Not really important, but try to format like black:
```cython
    if (
        mean_lower < mean < mean_upper
        and phasor_lower < real < phasor_upper
        and phasor_lower < imag < phasor_upper
    ):
```

src/phasorpy/phasor.py
```diff
@@ -2724,6 +2730,76 @@ def phasor_filter(
...
+    imag = numpy.asarray(imag)
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+
+    if out is None:
+        out = (mean.copy(), real.copy(), imag.copy())
+
+    return _mask_threshold(
+        *out, mean_lower, mean_upper, phasor_lower, phasor_upper, fill_value
+    )
```
cgohlke, line 2800:
```python
    return _mask_threshold(
        mean_lower, mean_upper, phasor_lower, phasor_upper, fill_value, **kwargs
    )
```

src/phasorpy/phasor.py
```diff
@@ -2724,6 +2730,76 @@ def phasor_filter(
...
+    """
+    mean = numpy.asarray(mean)
+    real = numpy.asarray(real)
+    imag = numpy.asarray(imag)
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+
+    if out is None:
+        out = (mean.copy(), real.copy(), imag.copy())
+
```
cgohlke, line 2797: None of this is necessary. It is handled by the ufunc.

cgohlke:
Hi @bruno-pannunzio: when you finished implementing the threshold function, could you also please update the LFD workshop tutorial? There are several places where thresholding is used:

https://github.com/phasorpy/phasorpy/blob/3106ce79e60a2643bd94b4834f899e291ad96e3d/tutorials/phasorpy_lfd_workshop.py#L224-L232

https://github.com/phasorpy/phasorpy/blob/3106ce79e60a2643bd94b4834f899e291ad96e3d/tutorials/phasorpy_lfd_workshop.py#L423-L429

https://github.com/phasorpy/phasorpy/blob/3106ce79e60a2643bd94b4834f899e291ad96e3d/tutorials/phasorpy_lfd_workshop.py#L475-L477

bruno-pannunzio:
> There should be no need to handle the `out` parameter. Just forward any `**kwargs` to the ufunc for example like in [`phasor_multiply`](https://github.com/phasorpy/phasorpy/blob/3106ce79e60a2643bd94b4834f899e291ad96e3d/src/phasorpy/phasor.py#L850-L908)

Thanks @cgohlke ! That's a relief. I thought it was going to be much difficult as I didn't think of the keyword arguments passed to the ufuncs. That works perfect now. Also thanks for you other comments, I will update with your suggestions. 

> could you also please update the LFD workshop tutorial?

Sure! I will update it in the next PR update.

> How about making the function as flexible as possible and have `_lower` and `_upper` parameters for `mean`, `real`, `imag`, `phase`, and `modulation`? That should be straightforward to implement in one ufunc, but if that does not work well (speed?) there could be simpler ufuncs for common cases that get dispatched in the main function. Currently `fill_value` stands out, because I don't see a need to ever varying values in one function call.

I am a little bit confused with this. How do you think the function signature should be? Also, do you think is better for the parameter names to use `_lower` and `_upper` rather than `_min` and `_max`?

cgohlke:
> How do you think the function signature should be?

```python
def phasor_threshold(
    mean: ArrayLike,
    real: ArrayLike,
    imag: ArrayLike,
    /,
    mean_lower: ArrayLike = -numpy.inf,
    mean_upper: ArrayLike = numpy.inf,
    *,
    real_lower: ArrayLike = -numpy.inf,
    real_upper: ArrayLike = numpy.inf,
    imag_lower: ArrayLike = -numpy.inf,
    imag_upper: ArrayLike = numpy.inf,
    phase_lower: ArrayLike = -numpy.inf,
    phase_upper: ArrayLike = numpy.inf,
    modulation_lower: ArrayLike = -numpy.inf,
    modulation_upper: ArrayLike = numpy.inf,
    fill_value: float = numpy.nan,
    **kwargs: Any,
 -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
```
Or is that too much?

Do you have a use case for `fill_value`?

> do you think is better for the parameter names to use _lower and _upper rather than _min and _max?

For a function named `phasor_threshold`, `lower` and `upper` feel more natural, but `min` and `max` are also good and shorter.

bruno-pannunzio:
> For a function named `phasor_threshold`, `lower` and `upper` feel more natural, but `min` and `max` are also good and shorter.

I agree. Maybe if we have that many parameters it's a good idea to keep them as short as possible. 

> Do you have a use case for `fill_value`?

I was thinking that the function can also be used as a filter and replacement of values. I don't know the exact use of now. But lets say someone needs to assign a range of values to 0 or other value for some reason, this function could do that as it is just changing the value to be replaced in the masked values. Also, maybe for some reason the `NaN` is incompatible with other functions outside the library, then it can be substituted by other value instead. But if you believe this function should not be used for that purpose to keep it simple, I also agree to that. 



> ```python
> def phasor_threshold(
>     mean: ArrayLike,
>     real: ArrayLike,
>     imag: ArrayLike,
>     /,
>     mean_lower: ArrayLike = -numpy.inf,
>     mean_upper: ArrayLike = numpy.inf,
>     *,
>     real_lower: ArrayLike = -numpy.inf,
>     real_upper: ArrayLike = numpy.inf,
>     imag_lower: ArrayLike = -numpy.inf,
>     imag_upper: ArrayLike = numpy.inf,
>     phase_lower: ArrayLike = -numpy.inf,
>     phase_upper: ArrayLike = numpy.inf,
>     modulation_lower: ArrayLike = -numpy.inf,
>     modulation_upper: ArrayLike = numpy.inf,
>     fill_value: float = numpy.nan,
>     **kwargs: Any,
>  -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
> ```

I think it has a lot of parameters, but it makes sense if we want to keep the function open for thresholding different values depending on the array. One alternative would be to have the range passed as a tuple (lower, upper) for each input, but I don't know if this is as intuitive as what you propose. 

Also, I was thinking, should the `phasor_threshold` function also accept `phase` and `modulation`? Although the logic would be exactly the same, I think if we want to be consistent with other functions of the module, separating phasor and polar processing should be done here, although it would just be repeating the same function with different parameter names as the logic is the same, right? If not we can rename the parameters to `real_or_phase` and `imag_or_modulation` but the lower and upper bound parameter names would be very long, and also the `phasor` in the name of the function can be confusing.

cgohlke:
> One alternative would be to have the range passed as a tuple (lower, upper) for each input, but I don't know if this is as intuitive as what you propose.

Better not. You want the function behave like an ufunc. Therefore, make the `fill_value` also array like (`fill_value: ArrayLike = numpy.nan`)?

> Also, I was thinking, should the phasor_threshold function also accept phase and modulation? 

Not the `phasor_threshold` function. I think the main use of the function will be right after `phasor_from_signal` (or any other user source of phasor coordinates) to get the phasor coordinates into a defined state. Unless you want to unify thresholding with the cursors, which are usually used later in the workflow?

I'm not sure, but it could make sense to set the default limits to be specific to phasor coordinates:

```Python
    real_lower: ArrayLike = -1.0,
    real_upper: ArrayLike = 1.0,
    imag_lower: ArrayLike = -1.0,
    imag_upper: ArrayLike = 1.0,
    phase_lower: ArrayLike = -2*numpy.pi,
    phase_upper: ArrayLike = 2*numpy.pi,
    modulation_lower: ArrayLike = 0.0,
    modulation_upper: ArrayLike = 1.0,
```

On the other hand, testing for phase limits involves calling trigonometric function, which are very slow. The comparison for limits should be ordered by how common they are (test the mean limits first) and effort (test the phase limits last).

cgohlke:
> do you think is better for the parameter names to use _lower and _upper rather than _min and _max?

The [cursors.mask_from_polar_cursor](https://www.phasorpy.org/stable/api/cursors.html#phasorpy.cursors.mask_from_polar_cursor) function uses `_min` and `_max`.

bruno-pannunzio:
@cgohlke I was thinking of the possibility of separating the function into two: `phasor_threshold` and `polar_threshold`. 

```python
def phasor_threshold(
    mean: ArrayLike,
    real: ArrayLike,
    imag: ArrayLike,
    /,
    mean_min: ArrayLike = -numpy.inf,
    mean_max: ArrayLike = numpy.inf,
    *,
    real_min: ArrayLike = -1.0,
    real_max: ArrayLike = 1.0,
    imag_min: ArrayLike = -1.0,
    imag_max: ArrayLike = 1.0,
    fill_value: ArrayLike = numpy.nan,
    **kwargs: Any,
 -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
```
and

```python
def phasor_polar_threshold(
    mean: ArrayLike,
    real: ArrayLike,
    imag: ArrayLike,
    /,
    mean_min: ArrayLike = -numpy.inf,
    mean_max: ArrayLike = numpy.inf,
    *,
    phase_min: ArrayLike = -numpy.pi,
    phase_max: ArrayLike = numpy.pi,
    modulation_min: ArrayLike = 0.0,
    modulation_max: ArrayLike = 1.0,
    fill_value: ArrayLike = numpy.nan,
    **kwargs: Any,
 -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
```

My reasoning is that with the previous implementation considering phasor and polar limits, if a value is not thresholded by the `mean` or phasor limits, it needs to be converted to polar coordinates to check for limits. This may be a little bit inefficient, considering most of the image is not supposed to be thresholded. With two separate functions, the conversion isn't done unless needed. The drawback I see here is that if you want to apply all thresholds (mean, phasor and polar) you need to call two separate functions, while the other way you can call it once, with all the impact this has in memory and execution performance. 

One other alternative I can think of is set the `_min` and `max_` from `phase` and `modulation` to `None` by default, and check if those values are different than `None` then perform the conversion and limit check. 

What do you think could be the best approach?

cgohlke:
> One other alternative I can think of is set the _min and _max from phase and modulation to None by default, and check if those values are different than None then perform the conversion and limit check.

I prefer this. Set all limits to None by default and then skip checks accordingly. It is sort of what I meant earlier: "That should be straightforward to implement in one ufunc, but if that does not work well (speed?) there could be simpler ufuncs for common cases that get dispatched in the main function." Or maybe it is enough to check in the general ufunc, which limits are set, maybe by passing NaN for no limit?

cgohlke:
Hi @bruno-pannunzio 

Here's what I had in mind (not tested). It's very pedestrian but straightforward, no surprises, and keeps the ufunc behavior. The function behaves a little different than what we discussed, not applying any threshold by default unless set (this could be changed), and rejecting values outside a closed interval (like `numpy.clip`). I'm not sure it's worth dispatching to specialized ufuncs. I think we should just write a simple benchmark comparing the runtimes also compared to the pure numpy way.

```Cython
@cython.ufunc
cdef (double, double, double) _phasor_threshold(
    float_t mean,
    float_t real,
    float_t imag,
    float_t mean_min,
    float_t mean_max,
    float_t real_min,
    float_t real_max,
    float_t imag_min,
    float_t imag_max,
    float_t phase_min,
    float_t phase_max,
    float_t modulation_min,
    float_t modulation_max,
    float_t fill,
) noexcept nogil:
    """Return thresholded values."""
    cdef:
        double phi = NAN
        double mod = NAN

    if isnan(mean) or isnan(real) or isnan(imag):
        return <float_t> NAN, <float_t> NAN, <float_t> NAN

    if not isnan(mean_min) and mean < mean_min:
        return fill, fill, fill
    if not isnan(mean_max) and mean > mean_max:
        return fill, fill, fill

    if not isnan(real_min) and real < real_min:
        return fill, fill, fill
    if not isnan(real_max) and real > real_max:
        return fill, fill, fill

    if not isnan(imag_min) and imag < imag_min:
        return fill, fill, fill
    if not isnan(imag_max) and imag > imag_max:
        return fill, fill, fill

    if not isnan(modulation_min):
        mod = hypot(real, imag)
        if mod < modulation_min:
            return fill, fill, fill
    if not isnan(modulation_max):
        if isnan(mod):
            mod = hypot(real, imag)
        if mod > modulation_max:
            return fill, fill, fill

    if not isnan(phase_min):
        phi = atan2(imag, real)
        if phi < phase_min:
            return fill, fill, fill
    if not isnan(phase_max):
        if isnan(phi):
            phi = atan2(imag, real)
        if phi > phase_max:
            return fill, fill, fill

    return mean, real, imag


@cython.ufunc
cdef (double, double, double) _phasor_threshold_mean(
    float_t mean,
    float_t real,
    float_t imag,
    float_t mean_min,
    float_t mean_max,
    float_t fill,
) noexcept nogil:
    """Return thresholded values."""

    if isnan(mean) or isnan(real) or isnan(imag):
        return <float_t> NAN, <float_t> NAN, <float_t> NAN

    if not isnan(mean_min) and mean < mean_min:
        return fill, fill, fill
    if not isnan(mean_max) and mean > mean_max:
        return fill, fill, fill

    return mean, real, imag
```
```Python
def phasor_threshold(
    mean: ArrayLike,
    real: ArrayLike,
    imag: ArrayLike,
    /,
    mean_min: ArrayLike | None = None,
    mean_max: ArrayLike | None = None,
    *,
    real_min: ArrayLike | None = None,
    real_max: ArrayLike | None = None,
    imag_min: ArrayLike | None = None,
    imag_max: ArrayLike | None = None,
    phase_min: ArrayLike | None = None,
    phase_max: ArrayLike | None = None,
    modulation_min: ArrayLike | None = None,
    modulation_max: ArrayLike | None = None,
    fill_value: ArrayLike | None = None,
    **kwargs: Any,
) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
    """Return phasor coordinates with values out of interval replaced by NaN.

    Interval thresholds can be set for mean intensity, real and imaginary
    coordinates, and phase and modulation.
    Phasor coordinates smaller than minimum thresholds or larger than maximum
    thresholds will be replaced by a fill value, NaN by default.
    No threshold is applied by default, all intervals are open.

    Parameters
    ----------
    mean : array_like
        Mean intensity of phasor coordinates.
    real : array_like
        Real component of phasor coordinates.
    imag : array_like
        Imaginary component of phasor coordinates.
    mean_min : array_like, optional
        Lower threshold for mean intensity.
    mean_max : array_like, optional
        Upper threshold for mean intensity.
    real_min : array_like, optional
        Lower threshold for real coordinates.
    real_max : array_like, optional
        Upper threshold for real coordinates.
    imag_min : array_like, optional
        Lower threshold for imaginary coordinates.
    imag_max : array_like, optional
        Upper threshold for imaginary coordinates.
    phase_min : array_like, optional
        Lower threshold for phase angle.
    phase_max : array_like, optional
        Upper threshold for phase angle.
    modulation_min : array_like, optional
        Lower threshold for modulation.
    modulation_max : array_like, optional
        Upper threshold for modulation.
    fill_value : array_like, optional
        Value used to fill out of interval elements.
    **kwargs
        Optional `arguments passed to numpy universal functions
        <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.

    Returns
    -------
    mean : ndarray
        Thresholded mean intensity of phasor coordinates.
    real : ndarray
        Thresholded real component of phasor coordinates.
    imag : ndarray
        Thresholded imaginary component of phasor coordinates.

    Examples
    --------
    Set phasor coordinates to NaN if mean intensity is smaller than 1.1:

    >>> phasor_threshold([1, 2, 3], [0.1, 0.2, 0.3], [0.4, 0.5, 0.6], 1.1)
    (array([nan, 2, 3]), array([nan, 0.2, 0.3]), array([nan, 0.5, 0.6]))

    Set phasor coordinates to NaN if real component is smaller than 0.15 or
    larger than 0.25:

    >>> phasor_threshold(
    ...     [1.0, 2.0, 3.0],
    ...     [0.1, 0.2, 0.3],
    ...     [0.4, 0.5, 0.6],
    ...     real_min=0.15,
    ...     real_max=0.25,
    ... )
    (array([nan, 2, nan]), array([nan, 0.2, nan]), array([nan, 0.5, nan]))

    """
    threshold_mean_only = None
    if mean_min is None:
        mean_min = numpy.nan
        threshold_mean_only = True
    if mean_max is None:
        mean_max = numpy.nan
        threshold_mean_only = True
    if real_min is None:
        real_min = numpy.nan
        threshold_mean_only = False
    if real_max is None:
        real_max = numpy.nan
        threshold_mean_only = False
    if imag_min is None:
        imag_min = numpy.nan
        threshold_mean_only = False
    if imag_max is None:
        imag_max = numpy.nan
        threshold_mean_only = False
    if phase_min is None:
        phase_min = numpy.nan
        threshold_mean_only = False
    if phase_max is None:
        phase_max = numpy.nan
        threshold_mean_only = False
    if modulation_min is None:
        modulation_min = numpy.nan
        threshold_mean_only = False
    if modulation_max is None:
        modulation_max = numpy.nan
        threshold_mean_only = False
    if fill_value is None:
        fill_value = numpy.nan
        threshold_mean_only = False

    if threshold_mean_only is None:
        return (
            numpy.asanyarray(mean),
            numpy.asanyarray(real),
            numpy.asanyarray(imag),
        )

    if threshold_mean_only:
        return _phasor_threshold_mean(
            mean, real, imag, mean_min, mean_max, fill_value
        )

    return _phasor_threshold(
        mean,
        real,
        imag,
        mean_min,
        mean_max,
        real_min,
        real_max,
        imag_min,
        imag_max,
        phase_min,
        phase_max,
        modulation_min,
        modulation_max,
        fill_value,
    )
```

cgohlke:
Here are some simple timings for three 1 GB arrays, filtering about 10% or none:
```Python
import numpy
from tifffile import Timer

from phasorpy._phasorpy import _phasor_threshold, _phasor_threshold_nan
from phasorpy.phasor import phasor_from_signal, phasor_threshold

mean, real, imag = numpy.random.random((3, 512, 512, 512))

mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('_phasor_threshold_nan'):
    mean1, real1, imag1 = _phasor_threshold_nan(mean1, real1, imag1)


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('_phasor_threshold_nan inplace'):
    _phasor_threshold_nan(mean1, real1, imag1, out=(mean1, real1, imag1))


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('_phasor_threshold all nan'):
    mean1, real1, imag1 = _phasor_threshold(
        mean1, real1, imag1, *((numpy.nan,) * 11)
    )


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('_phasor_threshold all nan inplace'):
    _phasor_threshold(
        mean1, real1, imag1, *((numpy.nan,) * 11), out=(mean1, real1, imag1)
    )


with Timer('numpy mean'):
    mean1 = mean.copy()
    real1 = real.copy()
    imag1 = imag.copy()
    mask = (mean < 0.05) | (mean > 0.95)
    mean1[mask] = numpy.nan
    real1[mask] = numpy.nan
    imag1[mask] = numpy.nan


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('numpy mean inplace'):
    mask = (mean1 < 0.05) | (mean1 > 0.95)
    mean1[mask] = numpy.nan
    real1[mask] = numpy.nan
    imag1[mask] = numpy.nan


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('mean'):
    mean1, real1, imag1 = phasor_threshold(mean1, real1, imag1, 0.05, 0.95)


mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('mean inplace'):
    phasor_threshold(
        mean1, real1, imag1, 0.05, 0.95, out=(mean1, real1, imag1)
    )

mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('phasor'):
    mean1, real1, imag1 = phasor_threshold(
        mean1, real1, imag1, real_min=0.05, imag_max=0.95
    )

mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('phasor inplace'):
    phasor_threshold(
        mean1,
        real1,
        imag1,
        real_min=0.05,
        imag_max=0.95,
        out=(mean1, real1, imag1),
    )

mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('polar'):
    mean1, real1, imag1 = phasor_threshold(
        mean1, real1, imag1, phase_min=numpy.pi / 10, modulation_max=0.95
    )

mean1 = mean.copy()
real1 = real.copy()
imag1 = imag.copy()
with Timer('polar inplace'):
    phasor_threshold(
        mean,
        real,
        imag,
        phase_min=numpy.pi / 10,
        modulation_max=0.95,
        out=(mean1, real1, imag1),
    )

```
```
_phasor_threshold_nan 0.550355 s
_phasor_threshold_nan inplace 0.167053 s
_phasor_threshold all nan 1.533958 s
_phasor_threshold all nan inplace 0.984733 s
numpy mean 1.382135 s
numpy mean inplace 0.859516 s
mean 0.790464 s
mean inplace 0.402861 s
phasor 1.511467 s
phasor inplace 0.972152 s
polar 3.382267 s
polar inplace 3.045243 s
```
Conclusions:

- thresholding takes significant time. Comparing just all NaNs inplace take ~1s
- thresholding just the intensity is almost twice as fast as in pure numpy
- thresholding polar coordinates is slow, as expected
- inplace operation are significantly faster, as expected

bruno-pannunzio:
Excelent @cgohlke! I think this works great. As you mentioned, the improvement in performance is nice. I also like the way the code it's presented, as you said it may be simple but that it also easier and more straightforward to understand what it is doing. I encountered the issue of Cython not working with None a problem, but with your alternative to passing `NaN` that problem is solved.

I will add some tests and I will set it up for review so that it can be merged. Thanks for the suggestions again!

src/phasorpy/_phasorpy.pyx
```diff
@@ -1512,3 +1512,105 @@ cdef float_t _blend_lighten(
...
+    float_t modulation_max,
+    float_t fill,
+) noexcept nogil:
+    """Return thresholded values."""
+    cdef:
+        double phi = NAN
+        double mod = NAN
+
+    if isnan(mean) or isnan(real) or isnan(imag):
+        return <float_t> NAN, <float_t> NAN, <float_t> NAN
```
cgohlke, line 1543: Should this return `fill` instead? On the other hand, I'm also not sure when one would want to set the intensity mean and phasor coordinates to the same value other than NaN or maybe zero. Zero could be risky as we have discussed, and the user could easily use [numpy.nan_to_num](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html) to replace NaNs with anything.

src/phasorpy/phasor.py
```diff
@@ -2688,6 +2693,174 @@ def phasor_filter(
...
+    modulation_max: ArrayLike | None = None,
+    fill_value: ArrayLike | None = None,
+    **kwargs: Any,
+) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
+    """Return phasor coordinates with values out of interval replaced by NaN.
+
+    Interval thresholds can be set for mean intensity, real and imaginary
+    coordinates, and phase and modulation.
+    Phasor coordinates smaller than minimum thresholds or larger than maximum
+    thresholds will be replaced by a fill value, NaN by default.
```
cgohlke, line 2720: will be -> are

tests/test_phasor.py
```diff
@@ -2155,3 +2156,162 @@ def test_phasor_filter_errors():
...
         phasor_filter([[0]], [[0]], repeat=-3)  # repeat < 1
+
+
+def test_phasor_threshold():
+    """Test `phasor_threshold` function."""
+    nan = numpy.nan
+    assert_allclose(
+        phasor_threshold([0.5, 0.4], [0.2, 0.5], [0.3, 0.5]),
+        ([0.5, 0.4], [0.2, 0.5], [0.3, 0.5]),
+    )  # no threshold
```
cgohlke, line 2167: I would prefer the comments in the line before rather than "inline"

tutorials/phasorpy_introduction.py
```diff
@@ -168,10 +168,9 @@
 # Pixels with low intensities are commonly excluded from analysis and
 # visualization of phasor coordinates:
 
-# TODO: replace this with a ``phasor_mask`` function?
-mask = mean > 1
-real = real[mask]
-imag = imag[mask]
+from phasorpy.phasor import phasor_threshold
+
+mean, real, imag = phasor_threshold(mean, real, imag, 1)
```
cgohlke, line 173: Probably doesn't matter, but I think the equivalent to filter `mean>1` is `phasor_threshold(mean, real, imag, 2)`, no?

tutorials/phasorpy_lfd_workshop.py
```diff
@@ -220,17 +221,13 @@
...
-imag1 = imag1[threshold]
-real2 = real2[threshold]
-imag2 = imag2[threshold]
 
+_, real1, imag1 = phasor_threshold(
+    mean, real1, imag1, 20, real_min=0, imag_min=0
+)
+_, real2, imag2 = phasor_threshold(
+    mean, real2, imag2, 20, real_min=0, imag_min=0
+)
```
cgohlke, line 230: This is OK, although again technically not the same. How should filtering for example `mean <= 0.0` instead of `mean < 0.0` be handled? Maybe `mean < 1e-9` is good enough?

Also, is there a way to put this in a single statement with broadcasting? Just wondering.

cgohlke:
Hi @bruno-pannunzio : This looks good. I left some comments, mostly for items to be aware of and/or for discussion. 

After this is merged, all functions taking phasor coordinates should be validated to work with NaNs. Some statements in the tutorials raise warnings.

src/phasorpy/_phasorpy.pyx
```diff
@@ -1512,3 +1512,105 @@ cdef float_t _blend_lighten(
...
+        return fill, fill, fill
+
+    if not isnan(modulation_min):
+        mod = hypot(real, imag)
+        if mod < modulation_min:
+            return fill, fill, fill
+    if not isnan(modulation_max):
+        if isnan(mod):
+            mod = hypot(real, imag)
+        if mod > modulation_max:
```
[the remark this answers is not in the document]
cgohlke: It seems to be 10% faster, so let's compare squares.

src/phasorpy/_phasorpy.pyx
bruno-pannunzio: Good question. I'm not sure, but I think is better to propagate the `NaN` rather than the fill value. Although if that is the case it should be documented somewhere in the docstring. You are also right that replacing the `NaN`s  should be easy. Maybe at this stage is better to remove the `fill_value` as you said and see if it is necessary in the future. I will update and fill with `NaN`.

tutorials/phasorpy_lfd_workshop.py
bruno-pannunzio: > This is OK, although again technically not the same. How should filtering for example `mean <= 0.0` instead of `mean < 0.0` be handled?

This is a very good question also. Before, I had implemented it so that the lower bound is included, but I just realized that the new implementation doesn't consider the bounds. I think it's better to include the bounds (similar to what happens with the masking of different shapes you implemented, but in that case it is only included the lower bound). I think it also makes sense considering the name of the parameters, meaning that `_min` is the minimum value that is not thresholded. I don't know how other libraries that apply threshold in imaging handle this, but I will research and maybe we can discuss the better approach further. 


> Also, is there a way to put this in a single statement with broadcasting? Just wondering.

I left it like this since to try to keep the same structure as before, but I will look into it.

cgohlke: > I left it like this since to try to keep the same structure as before, but I will look into it.

I am just curious since multiple harmonics are returned by `phasor_from_signal` in one array.

Re bounds: There are certainly arguments for both <=, >= vs <, >. My reference was to keep it similar to `numpy.clip`, which looks a lot like our function. Also, we consider the phasor coordinates 0.0 and 1.0 valid, while > 1.0 and < 0.0 (for calibrated lifetimes) are not, and the function is probably more often used to remove "invalid" values. If needed, we could add a switch like `open_interval: bool = ...` and have two sets of Cython functions?

cgohlke: > I think it's better to include the bounds (similar to what happens with the masking of different shapes you implemented, but in that case it is only included the lower bound

Those functions seem consistent with what is implemented here. Values outside the closed interval [min..max] are rejected (False or NaN). No? With the exception of the range function. The argument for range to be different from interval is that (1) Python range excludes the upper value and (2) several ranges may be stacked together. Although we may want to review if that makes sense for floating point.

bruno-pannunzio: Those are all good points you raised Christoph. I think every approach has it's advantages and disadvantages, but maybe the best is to add the `open_interval` to make as flexible as possible without adding that much complexity?

> we consider the phasor coordinates 0.0 and 1.0 valid, while > 1.0 and < 0.0 (for calibrated lifetimes) are not, and the function is probably more often used to remove "invalid" values.

Also considering this, it makes a lot of sense to set by default to include the bounds, as not to exclude the 1.0 and 0.0. Do you think this can be the best approach?

bruno-pannunzio: Sorry @cgohlke I misunderstood the approach I think. You are right, the current implementation considers the closed interval (the default condition), but maybe it would be nice to add the option to make it open?

cgohlke: Maybe I am missing something, but the current implementation does not exclude min or max. It returns NaN only if a value `< min` or `> max`. No?

bruno-pannunzio: Exactly! I was wrong and I thought it was the other way. But it actually does that currently. Should we keep like this or add the `open_interval`?

cgohlke: > Should we keep like this or add the open_interval?

I leave that up to you. It's straightforward to implement but needs to be tested and documented too.

cgohlke:
> Let me try to verify this first.

Numpy is smarter: the function uses single precision if arrays are numpy.float32 and scalars are Python float. That is perfect for us.

bruno-pannunzio:
Great @cgohlke! So should we change the `_min` and `_max` to double instead or keep it as it is for now?

cgohlke:
> So should we change the _min and _max to double instead or keep it as it is for now?

No, everything is good as it is now.

src/phasorpy/phasor.py
```diff
@@ -2709,15 +2711,15 @@ def phasor_threshold(
...
     **kwargs: Any,
 ) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
     """Return phasor coordinates with values out of interval replaced by NaN.
 
     Interval thresholds can be set for mean intensity, real and imaginary
     coordinates, and phase and modulation.
     Phasor coordinates smaller than minimum thresholds or larger than maximum
-    thresholds will be replaced by a fill value, NaN by default.
+    thresholds are replaced NaN.
     No threshold is applied by default, all intervals are open.
```
cgohlke, line 2723: ", all intervals are open." can be removed now.

src/phasorpy/phasor.py
```diff
@@ -2831,18 +2836,37 @@ def phasor_threshold(
...
+            real_max,
+            imag_min,
+            imag_max,
+            phase_min,
+            phase_max,
+            modulation_min,
+            modulation_max,
+            **kwargs,
+        )
+    return _phasor_threshold_closed(
```
cgohlke, line 2869: In case you want to shorten this:
```Python
func = _phasor_threshold_open if open_interval else _phasor_threshold_closed
func(
    mean,
    real,
    imag,
    mean_min,
    ...
)
```

cgohlke (approved):
Looks good. I left two small comments.

bruno-pannunzio:
Thanks @cgohlke for those final suggestions. I will look into the compatibility of `NaN` with other functions in the library after merging.

cgohlke:
> I will look into the compatibility of NaN with other functions in the library after merging.

I was about to look into that, but if you'd like to follow up that's OK. Several Cython functions need to be guarded for NaNs in phasor and polar coordinates, and apparent lifetimes.

It might make sense to test for compatibility with NaNs in a separate test module, for example, `test_nan.py` ?

bruno-pannunzio:
Great, I think it makes more sense for you to look into the Cython functions since you implemented them. I will take a look and maybe if I find something I can start an issue and we see how we can work it out. 

> It might make sense to test for compatibility with NaNs in a separate test module, for example, `test_nan.py` ?

Totally! It's a good idea to gather `NaN` testing together for clarity and also to make sure everything is taken into consideration.

cgohlke:
Actually, the `open_interval` parameter turns out to be really useful when dealing with phasor coordinates that had NaNs replaced with zero (like apparently SimFCS ref formats).

bruno-pannunzio:
Hi @cgohlke! As I was working on the tutorial for multi-harmonic calibration I run into quite a problem regarding `phasor_threshold`.  Since the function is based on the cython ufuncs, the broadcasting between `mean`, `real` and `imag` is done automatically. The problem is that `mean` usually has less dimensions than `real` and `imag`, at least in the case of multiple harmonics, and when using `phasor_threshold` in this case, the returned `mean` has an extra dimension which was not there, corresponding to the harminc axis in `real` and `imag`. I thought of reshaping the result mean to match the original but then I run into something that needed further discussion in my opinion:

As it is now, the `NaN` are propagated between `mean`, `real` and `imag`. In the case that `mean` is reshaped to match the original, we need to decide what to do with the `NaN` propagated from different harmonics of `real` and `imag`. I can think of two options:
1. If any equivalent vale in any harmonic of `real` and `imag` are `NaN` then that is propagated to all harmonics and to the `mean`. 
2.  Don't propagate `NaN` from `real` and `imag` to `mean`, which can be tricky

A third alternative would be to have different `mean` for each harmonic but it doesn't make much sense in my opinion. Maybe there is another way of handlng this and avoid the reshaping, but I think the same issue with the `NaN` will still be there.

cgohlke:
So far we filter and threshold multiple harmonics separate, which is not really what we want. For example:

https://github.com/phasorpy/phasorpy/blob/c477ffc02d27f60b5e0a31101a5d1656350963f5/tutorials/phasorpy_lfd_workshop.py#L221-L225

The `phasor_threshold` function is intended to propagate any NaN to all harmonics/components (mean, real1, imag1, real2, imag2, etc). That would be your option one. No? Do you have any code to review that demonstrates the issue?

bruno-pannunzio:
> So far we filter and threshold multiple harmonics separate, which is not really what we want. For example:

Exactly, I think they should work for all harmonics at once.

Right now the `phasor_threshold` function works like this:

```Python

from phasorpy.phasor import phasor_threshold
import numpy

mean = numpy.ones((3, 3))
real = numpy.ones((2, 3, 3))
imag = real.copy()
real[0, 1, 1] = numpy.nan
mean, real, imag = phasor_threshold(mean, real, imag, 1)
print(mean)
```
```
[[[ 1.  1.  1.]
  [ 1. nan  1.]
  [ 1.  1.  1.]]

 [[ 1.  1.  1.]
  [ 1.  1.  1.]
  [ 1.  1.  1.]]]
```

So it modifies also `mean` to have the harmonic axis and the `NaN` are propagated to each harmonic separately. Do you think we should modify to propagate to all harmonics also? In the case of the `mean` it should do something like squashing all harmonics into one, which will only set as `NaN` wherever there is a `NaN` in any harmonic.

cgohlke:
> Do you think we should modify to propagate to all harmonics also?  In the case of the mean it should do something like squashing all harmonics into one, which will only set as NaN wherever there is a NaN in any harmonic.

Yes, that's what I meant. Like a single mask calculated from various parameters, applied to all phasor components (including mean and harmonics).

The harmonic axis would need to be treated specially in `phasor_threshold`. These are reduction operations, which ufuncs cannot do and I don't think there's any special support in Cython. As mentioned before, if supporting multiple harmonics is too complicated, maybe it could be implemented in a separate, specialized function, either with Python and numpy, or in a specialized Cython function (like _phasor_from_lifetime or _phasor_from_lifetime). 

Do you think this needs to be done now, before the first release? We have so many things piling up.

bruno-pannunzio:
> Do you think this needs to be done now, before the first release? We have so many things piling up.

You are right, I think this might take some time to implement the changes, so it may be a good idea to leave it in the backlog for future improvements. I will open an issue to allow tracking the problem and make the tutorial similar to the LFD tutorial, using harmonics separately.

Issue #47, opened 2024-03-19
Multiple component analysis


bruno-pannunzio:
This issue proposes the addition of a `phasorpy.components` module for the analysis and quantification of multiple components, both for fluorescence lifetime and spectral imaging. The following list encompasses some, but not all, of the features useful for component analysis that are proposed to be implemented:

- [ ] Location of second unknown component when location of only one is known.
- [ ] The contribution of two components in a phasor point when the location of pure individual species is known.
- [ ] Contribution of three and four components in a phasor point when the location of pure individual species is known.
- [ ] Blind component analysis: method using more than one harmonic to resolve components and their fractional intensities when location of pure species is unknown.

PR #106, opened 2024-09-01
Add algebraic solution for multiple component analysis

schutyb:
## Description

This PR adds functions to perform multiple component analysis by solving algebraically. This functionality is discussed in #47 and #190. The two functions added to the `components` module for this purpose are:

- `n_fractions_from_phasor` for computing the fractions for each of the multiple components. Based on `scipy.linalg.lstsq` for solving the system of equations.
- `components_matrix` to create the coefficient matrix of the components that is used alongside the phasor coordinates of the unknown mixture in `scipy.linalg.lstsq`.

## Checklist


schutyb:
Hi @bruno-pannunzio I add this functions as we were talking before. I need some help: 

1 - The data is in another zenodo so I don´t know how to import it from that. 
2 - Try the example in the tutorial with the data Alex gave to us, and you will see the result for each fractions are inconsistentes, there can be some negative values that corresponds to those components that have much noise, but not in the entire data. 

Have a look whenever you can and let me know later.

cgohlke:
> data is in another zenodo
> the data Alex gave to us

Where can we find those data?

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+from phasorpy import phasor_from_signal
+
+# TODO add the data from https://zenodo.org/records/
+
+image = tifffile.imread(
+    "/Users/schutyb/Documents/Projects/phasorpy_in_use/data/"
+    "38_Hoechst_Golgi_Mito_Lyso_CellMAsk_404_488_561_633_SP.lsm"
+)
+
+avg, re, im = phasor_from_signal(image, axis=0, harmonic=[1, 2])
```
cgohlke, line 280: Use PhasorPy conventions:
```
mean, real, imag = phasor_from_signal(signal, axis=0, harmonic=[1, 2])
```

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+)
+
+matrixA = numpy.asarray(
+    [
+        [0.5899, 0.5807, 0.1510, -0.5963, -0.6268],
+        [0.2162, -0.6531, -0.2347, 0.6098, -0.8306],
+        [0.1599, -0.6015, -0.6611, -0.6188, 0.7233],
+        [-0.0586, -0.1255, 0.1321, 0.1578, 0.1225],
+        [1, 1, 1, 1, 1],
+    ]
```
cgohlke, line 364: Demonstrate or explain how to obtain these coefficient matrices.

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+)
+# la matrix A es [g1, g2, s1, s2, 1] con los indices 1 y 2 para los armonicos
+
+fractions = components_unmixing_from_phasor_2(
+    re.reshape(256, 256, 2), im.reshape(256, 256, 2), matrixA
+)
+# %%
```
cgohlke, line 371: Plot the fractions.

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+
+plot = PhasorPlot(allquadrants=True, title='First harmonic phasor')
+plot.hist2d(re[0], im[0], cmap='RdYlBu_r')
+
+plot = PhasorPlot(allquadrants=True, title='Second harmonic phasor')
+plot.hist2d(re[1], im[1], cmap='RdYlBu_r')
+plt.show()
+
+
+def components_unmixing_from_phasor_2(
```
cgohlke, line 290: Import the function from `phasorpy.components`?

src/phasorpy/components.py
```diff
@@ -311,3 +312,123 @@ def graphical_component_analysis(
...
+    -------
+    fractions : ndarray
+        Fractions of each components.
+
+    Raises
+    ------
+    ValueError
+        If multi_harmonic_real and multi_harmonic_imag have different shape.
+        If coefficiency matrix is empty.
+
```
cgohlke, line 344: Add a reference to publication :ref:`Vallmitjana et al 2022 <vallmitjana-2022>`?

src/phasorpy/components.py
```diff
@@ -311,3 +312,123 @@ def graphical_component_analysis(
...
+            for c in range(multi_harmonic_real.shape[1]):
+                vecB = (
+                    [multi_harmonic_real[r, c, j] for j in range(nh)]
+                    + [multi_harmonic_imag[r, c, j] for j in range(nh)]
+                    + [1]
+                )
+                fractions[r, c] = numpy.linalg.lstsq(
+                    matrixA, vecB, rcond=None
+                )[0]
+        return fractions
```
cgohlke, line 434: The function is not tested.

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
 # %%
 # sphinx_gallery_thumbnail_number = 5
 # mypy: disable-error-code="arg-type"
```
cgohlke, line 261: These lines should be last in the tutorial.

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+
+
+# %%
+# Components unmixing from phasor
+# -------------------------------
+#
+# Multi components unmixing from phasor.
```
cgohlke, line 268: Add a short description of the method and dataset.

src/phasorpy/components.py
```diff
@@ -311,3 +312,123 @@ def graphical_component_analysis(
...
+    ncomp = matrixA.shape[0]
+    nh = int(ncomp / 2)
+    if len(multi_harmonic_real.shape) == 1:
+        vecB = (
+            [multi_harmonic_real[j] for j in range(nh)]
+            + [multi_harmonic_imag[j] for j in range(nh)]
+            + [1]
+        )
+        return numpy.linalg.lstsq(matrixA, vecB, rcond=None)[0]
+    else:
```
cgohlke, line 420: No need for `else:`

src/phasorpy/components.py
```diff
@@ -311,3 +312,123 @@ def graphical_component_analysis(
...
+    ...     matrixA) # doctest: +NUMBER
+    array([[[0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279]],
+    ...        [[0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279]],
+    ...        [[0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279],
+    ...         [0.32982308, 0.11319236, 0.41102329, 0.14604279]]])
```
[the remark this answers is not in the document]
schutyb: hi @cgohlke ok with this just to make sure, I will move it to the unit test, but shoul I leave the example empty here?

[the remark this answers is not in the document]
cgohlke: The first example can probably be simplified. Drop the named arguments and shorten the values.

cgohlke:
Do I understand correctly that the harmonics are expected in the last dimensions of the phasor coordinates? 

When implementing `phasor_from_signal` function, we settled that harmonics, if any, are in the first dimension. Ideally, this is handled with an `axis` parameter but that would complicate the implementation of several functions. Anyway, the default for harmonics should be `axis=0`.

tutorials/api/phasorpy_components.py
```diff
@@ -258,3 +259,113 @@
...
+
+import tifffile
+from phasorpy import phasor_from_signal
+
+# TODO add the data from https://zenodo.org/records/
+
+image = tifffile.imread(
+    "/Users/schutyb/Documents/Projects/phasorpy_in_use/data/"
+    "38_Hoechst_Golgi_Mito_Lyso_CellMAsk_404_488_561_633_SP.lsm"
+)
```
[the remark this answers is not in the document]
cgohlke: The dataset was added in #109

schutyb:
> Do I understand correctly that the harmonics are expected in the last dimensions of the phasor coordinates?
> 
> When implementing `phasor_from_signal` function, we settled that harmonics, if any, are in the first dimension. Ideally, this is handled with an `axis` parameter but that would complicate the implementation of several functions. Anyway, the default for harmonics should be `axis=0`.

Yes, because when using the **numpy.linalg.lstsq(a, b)** [https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html](url) function I pass parametrer b, with the real and imag part for each hamonic there, that is the way I found to relate coefficients in the parametres a and b in **lstsq**, there maybe a better way. I did this because we need to apply **lstsq** for each (real, imag) pair in the image.

I will update the PR soon when I finish your review and corrections. And we can find a better way to solve the linear eqaution, 
do you know or suggets another alternativer to **numpy.linalg.lstsq(a,** b)? 

Thanks

schutyb:
> > Do I understand correctly that the harmonics are expected in the last dimensions of the phasor coordinates?
> > When implementing `phasor_from_signal` function, we settled that harmonics, if any, are in the first dimension. Ideally, this is handled with an `axis` parameter but that would complicate the implementation of several functions. Anyway, the default for harmonics should be `axis=0`.
> 
> Yes, because when using the **numpy.linalg.lstsq(a, b)** [https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html](url) function I pass parametrer b, with the real and imag part for each hamonic there, that is the way I found to relate coefficients in the parametres a and b in **lstsq**, there maybe a better way. I did this because we need to apply **lstsq** for each (real, imag) pair in the image.
> 
> I will update the PR soon when I finish your review and corrections. And we can find a better way to solve the linear eqaution, do you know or suggets another alternative to **numpy.linalg.lstsq(a,** b)?
> 
> Thanks

I have updated the PR but I haven´t finished yet, so it has many unchecks test, I let you know when its ready for review and pass all checks

cgohlke:
> because we need to apply lstsq for each (real, imag) pair in the image.

I understand that the inputs for `lstsq` need to be in a certain way. But you already generate those in manual loops. Just change the indexing from `[..., j]` to `[j, ...]`. No?

All PhasorPy functions should ideally work with N-dimensional arrays. This function seems hard-coded for two dimensions in addition to harmonic. How about using numpy iterators instead? Or better vectorize the loop(s) if possible.

Also make sure the function works with NaN input. If `numpy.linalg.lstsq` does not work with NaNs, maybe `scipy.linalg.lstsq` is a better choice?

schutyb:
Hi @cgohlke can you help me undertand the errors I got. 

I do not know why one of the checks for the example I wrote in the phasor_based_unmixing function gives a different output, also compare with the other checks. 

> 
> =================================== FAILURES ===================================
> _____________ [doctest] phasorpy.components.phasor_based_unmixing ______________
> 357         The array shapes of `real` and `imag`
> 358         The coefficient matrix is empty
> 359 
> 360     Example
> 361     -------
> 362     >>> real = numpy.array([0.5, 0.3])
> 363     >>> imag = numpy.array([0.2, 0.7])
> 364     >>> coeff_matrix = numpy.array([[0.5, 0.3], [0.2, 0.7], [1.0, 1.0]])
> 365 
> 366     >>> phasor_based_unmixing(
> Expected:
>     (0.9999999999999994, 4.229244120114301e-16)
> Got:
>     (0.9999999999999994, 3.9099874730191774e-16)

Then I do not undestand the **codecov/patch**  warning.

After I corect this I can ask you for review. 

Thanks

cgohlke:
> (0.9999999999999994, 3.9099874730191774e-16)

Those numbers are 1.0 and 0.0 within expected precision. Not a good example for docstrings. Maybe better to use `(1.0, 0.0)` with `# doctest: +SKIP`

The codecov warnings are pointing out lines of codes that are not covered by tests.

schutyb:
Hi @cgohlke it is ready for review

schutyb:
Perfect, I will do them now, and update.

 I completly forget about the R description since I have it open from months, my bad.

@bruno-pannunzio is leaving on monday I think for some week of and so am I, but I will return in late March, so if I can solve everything at leat to have a first version to merge before monday, I´ll try to impreve as much as I can with your review, to prevent Bruno from being overworked when he returns.

bruno-pannunzio:
Thanks @schutyb. As you said, I can work on were you left when I get back in two weeks. 

But I agree with Christoph comments. It is also a good opportunity to review the function names of the module. I like `n_fractions_from_phasor` for the function proposed here. Maybe we can rename `two_fractions_from_phasor` for something related to the method of projecting to the line, but I can't think of a better and short name. `fractions_from_line_between_two_components`? It may be too long and is not entirely correct. `two_fractions_from_line` also is not correct but may get to the point?

Also it would be nice to handle the calculation of the coefficient matrix internally so as not to leave it to the user. I think in line with other functions of the library, it can have as input the real and imaginary coordinates (separately) of the components instead.

schutyb:
Thanks @bruno-pannunzio 

> Also it would be nice to handle the calculation of the coefficient matrix internally so as not to leave it to the user. I think in line with other functions of the library, it can have as input the real and imaginary coordinates (separately) of the components instead.

I disagree with this because the coefficients matrix is done with the real and imag for each pure components, so in a experiment you take many images, if you apply the unmixing function many time for each image in you will create the matrix every time, instead we can just create the matrix one time and pass then everytime you call the function. 

But we can discuss later.

cgohlke:
That makes sense. In that case, mention the use case in the documentation/tutorial and provide a separate function that assembles the coefficient matrix from any number of component phasor coordinates. The idea is to shield the user from the internal structure of the coefficient matrix and to make it straightforward to create. Depending how complicated the interface is, you may want to make the `phasor_based_unmixing` function optionally take the coefficient matrix (for repeated use) or any number of component phasor coordinates (for single use).

schutyb:
> That makes sense. In that case, mention the use case in the documentation/tutorial and provide a separate function that assembles the coefficient matrix from any number of component phasor coordinates. The idea is to shield the user from the internal structure of the coefficient matrix and to make it straightforward to create. Depending how complicated the interface is, you may want to make the `phasor_based_unmixing` function optionally take the coefficient matrix (for repeated use) or any number of component phasor coordinates (for single use).

Perfect!

src/phasorpy/components.py
```diff
@@ -311,3 +315,209 @@ def graphical_component_analysis(
...
+
+
+def n_fractions_from_phasor(
+    real: ArrayLike,
+    imag: ArrayLike,
+    /,
+    *,
+    components_real: ArrayLike | None = None,
+    components_imag: ArrayLike | None = None,
+    components_matrix: NDArray[Any] | None = None,
```
cgohlke, line 327: I would prefer singular names (`component_real`, not `components_real`) and not needing to pass `component_real` and `component_imag` by name, for example:

```suggestion
    real: ArrayLike,
    imag: ArrayLike,
    /,
    component_real: ArrayLike | None = None,
    component_imag: ArrayLike | None = None,
    *,
    component_matrix: ArrayLike | None = None,
```

src/phasorpy/components.py
```diff
@@ -311,3 +315,209 @@ def graphical_component_analysis(
...
+    fractions = fractions.reshape((num_components,) + coords.shape[1:])
+    fractions = [numpy.where(invalid_mask, numpy.nan, f) for f in fractions]
+
+    return tuple(map(numpy.asarray, fractions))
+
+
+def components_matrix_from_phasor(
+    components_real: ArrayLike,
+    components_imag: ArrayLike,
+) -> NDArray[Any]:
```
cgohlke, line 456: How about naming this function just `component_matrix` or `phasor_component_matrix`? Make argument positional:

```suggestion
def phasor_component_matrix(
    component_real: ArrayLike,
    component_imag: ArrayLike,
    /,
) -> NDArray[Any]:
```

src/phasorpy/components.py
cgohlke: Never mind the singular names. I see that the other functions in the components module also use these names. It's more important to be consistent.

bruno-pannunzio: Hi @cgohlke, thanks for pointing this out, I also think is better to set this parameters as positional. Regarding the names, I also like the singular form, but since the components will always be more than one (since it doesn't make sense a one component analysis), I think it's more intuitive to leave it as plural. But if you think is better singular I can also make the modifications in the other functions of the module to be consistent with this one.

src/phasorpy/components.py
bruno-pannunzio: Thanks Christoph, this is a good suggestion. I will change this and push some other modifications I've been working on, also in the tutorial.

bruno-pannunzio:
@cgohlke I made some additional modifications to the function and an extensive revamp of the section of the components tutorial. I will leave some more comments in line for us to discuss when possible. The test are failing due to the changes, I will look at that later and also add more test since none of them are actually validating the numerical results at the moment.

tutorials/api/phasorpy_components.py
```diff
@@ -244,6 +252,142 @@
...
+    )
+
+    plot_h2.hist2d(real[1], imag[1], cmap='RdYlBu_r', bins=300)
+    plot_h2.plot(
+        components_real[1][i],
+        components_imag[1][i],
+        marker='o',
+        markersize=8,
+        label=components_names[i],
+    )
```
bruno-pannunzio, line 333: I think either this section of plotting both of the harmonics of the components or the samples should be removed and keep only one of them for showcasing. Although it looks nice, maybe the plotting of the pure components is not really needed in this case.

src/phasorpy/components.py
bruno-pannunzio: I renamed to `components_matrix` to be consistent with the previous discussion, but if we decide to go with singular I can modify this also.

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+    components_matrix: NDArray[Any] | None = None,
+    **kwargs: Any,
+) -> tuple[NDArray[Any], ...]:
+    """Return fractions of multiple components from phasor coordinates.
+
+    Parameters
+    ----------
+    real : array_like
+        Real component of phasor coordinates.
+        The harmonics must be in the first dimension.
```
bruno-pannunzio, line 151: This phrase can be improved for clarity, indicating that the harmonic must be present in the first axis (if present)?

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+    components_real: array_like, optional
+        Real coordinates of the components.
+        The harmonics must be in the first dimension.
+    components_imag: array_like, optional
+        Imaginary coordinates of the components.
+        The harmonics must be in the first dimension.
+    components_matrix : array_like, optional
+        Components coefficient matrix.
+        The shape of the matrix must be (2h+1, n), where `h` is
+        the number of harmonics and `n` is the number of components.
```
bruno-pannunzio, line 164: I think it is necessary to indicate the expected shape of the matrix, but I don't know if this is the best way to indicate how it should be.

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+        components_matrix = components_matrix(components_real, components_imag)
+
+    components_matrix = numpy.asarray(components_matrix)
+
+    num_harmonics = (components_matrix.shape[0] - 1) / 2
+    num_components = components_matrix.shape[1]
+
+    if components_matrix.shape[0] < num_components:
+        raise ValueError(
+            f'{components_matrix.shape[0]=} < {components_matrix.shape[1]=}'
```
bruno-pannunzio, line 229: This error is raised even though `scipy.linalg.lstsq` can solve even if the system is undetermined. It minimizes the norm and gives the solution with this minimized, although theoretically there can be infinite solutions. That is why I don't think is correct to allow this in the case of the component analysis. I don't think anybody has go into detail with this since the method is quite recent, but I don't know. The alternative would be to remove this check and leave it open to work even if the system is undetermined.

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+
+    invalid_mask = numpy.any(
+        numpy.isnan(real)
+        | numpy.isnan(imag)
+        | numpy.isinf(real)
+        | numpy.isinf(imag),
+        axis=0,
+    )
+    real = numpy.nan_to_num(real, nan=0.0, posinf=0.0, neginf=0.0)
+    imag = numpy.nan_to_num(imag, nan=0.0, posinf=0.0, neginf=0.0)
```
bruno-pannunzio, line 249: Since `scipy.linalg.lstsq` does not allow `NaN` of infinites, this should be taken care of before calling the function. I think the best way is to assign 0 and then restore the original invalid values. Do you think this is ok?

src/phasorpy/components.py
```diff
@@ -37,6 +45,7 @@
     from ._typing import Any, ArrayLike, NDArray
 
 import numpy
+import scipy
```
cgohlke, line 48: Delay import in the function?

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+
+    if components_matrix is None:
+        if components_real is None or components_imag is None:
+            raise ValueError(
+                'either the components matrix or the components phasor '
+                'coordinates must be provided'
+            )
+        components_matrix = components_matrix(components_real, components_imag)
```
cgohlke, line 220: The names of variable and function conflict. How about naming the function `phasor_component_matrix` so it can later be added to a flat phasorpy namespace?

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+
+    components_matrix = numpy.asarray(components_matrix)
```
cgohlke, line 222:
```suggestion
    else:
        components_matrix = numpy.asarray(components_matrix)
```

src/phasorpy/components.py
cgohlke: Let's leave this check at least for now.

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+    ones_array = numpy.ones((1,) + real.shape[1:])
+    coords = numpy.concatenate([real, imag, ones_array], axis=0)
+
+    fractions = scipy.linalg.lstsq(
+        components_matrix, coords.reshape(coords.shape[0], -1), **kwargs
+    )[0]
+    fractions = fractions.reshape((num_components,) + coords.shape[1:])
+    fractions = [numpy.where(invalid_mask, numpy.nan, f) for f in fractions]
+
+    return tuple(map(numpy.asarray, fractions))
```
cgohlke, line 260: Why is mapping to asarray necessary here?

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+    imag = numpy.nan_to_num(imag, nan=0.0, posinf=0.0, neginf=0.0)
+
+    ones_array = numpy.ones((1,) + real.shape[1:])
+    coords = numpy.concatenate([real, imag, ones_array], axis=0)
+
+    fractions = scipy.linalg.lstsq(
+        components_matrix, coords.reshape(coords.shape[0], -1), **kwargs
+    )[0]
+    fractions = fractions.reshape((num_components,) + coords.shape[1:])
+    fractions = [numpy.where(invalid_mask, numpy.nan, f) for f in fractions]
```
cgohlke, line 258: Does this pass mypy?

src/phasorpy/components.py
cgohlke: It's not only the shape of the matrix that needs to be described. It's probably enough to refer to the  `components_matrix` function and add details there.

src/phasorpy/components.py
cgohlke: These are expensive operations. Many arrays are created intermediately. It might help passing the `mean` coordinates to the function and/or Cythonizing these operations. This could be done later, after the tests and tutorial. 

Having access to the `mean` coordinates could also improve harmonic handling.

src/phasorpy/components.py
```diff
@@ -123,6 +132,208 @@ def two_fractions_from_phasor(
...
+def n_fractions_from_phasor(
+    real: ArrayLike,
+    imag: ArrayLike,
+    components_real: ArrayLike | None = None,
+    components_imag: ArrayLike | None = None,
+    /,
+    *,
+    components_matrix: NDArray[Any] | None = None,
+    **kwargs: Any,
+) -> tuple[NDArray[Any], ...]:
```
cgohlke, line 144: I think the function might benefit from adding `mean` and `harmonic` parameters, mainly to verify the presence of an harmonic axis and to verify the actual harmonics present in the phasor coordinates. `mean` could be used to mask NaN and Inf more efficiently (see below).

src/phasorpy/components.py
cgohlke: Sorry, I'm used to "component matrix", e.g. from PCA.

src/phasorpy/components.py
bruno-pannunzio: good idea!

src/phasorpy/components.py
bruno-pannunzio: You are right, I don't think it's needed. This was left from a previous implementation.

src/phasorpy/components.py
bruno-pannunzio: @cgohlke I think it's a good idea to add `mean` as parameter. As you said we can use to check for harmonic axis and also to propagate the `NaN`s and `Inf`s. I tried implementing this with `phasor_threshold` but I don't know if this is what you had in mind and if it is more efficient than before.

Regarding the `harmonic` parameter. I see that it can be good to check the number of harmonics present in the phasor coordinates, but other from that I don't see that the exact values of the harmonics be used elsewhere in the function. Am I missing some other use for the harmonic parameter here?

cgohlke: The optional `harmonic` parameter would be used for validation as a safeguard. I was thinking of the phasor coordinates returned by the `io` functions, which may not necessarily contain the lower harmonics.

bruno-pannunzio: Oh I see. But in this case what we need is how many harmonics are present, rather than which harmonics are present. As long as the same sequence of harmonics are in the phasor coordinates and the component matrix we are ok, but I don't think we can check this easily, just how many of them are.

cgohlke: OK. If the values of harmonics really don't matter then the parameter is not needed. I agree that it is enough to only verify the shape, not the values, of the component matrix if it is provided.

cgohlke:
Superseded by #238

Issue #87, opened 2024-08-01
Improve consistency of handling `NaN` by `phasor.phasor_filter` 


bruno-pannunzio:
Current implementation of median filter by `phasor.phasor_filter` uses `scipy.ndimage.median_filter` which has been [reported to have undefined behaviour with `NaN`](https://github.com/scipy/scipy/issues/4800). It has been [proposed to add options to have more control on `NaN`](https://github.com/scipy/scipy/issues/10807), but this was many years ago and it's still unanswered, so I am guessing they are not interested in implementing this changes for now.

One solution would be to implement a median filter independent from `scipy.ndimage.median_filter` or `skimage.filters.median`, since both are based in the same filter. This can be implemented with Cython, which can also help with performance, considering iterations of filtering are common with phasor coordinates filtering, so this can also be leveraged to improve the filtering perfomance, as well as having more control over `NaN` handling.

Another solution would be to replace the median filtering with another function that has more consistency for handling `NaN`.

cgohlke:
Closed by #147 and #150.

PR #147, opened 2024-11-20
Replace median filter implementation for NaN handling consistency

bruno-pannunzio:
## Description

This PR proposes the replacement of the current implementation of the median filter method in `phasorpy.phasor.phasor_filter`, which is a convenience wrapper around [`scipy.ndimage.median_filter`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.median_filter.html#scipy.ndimage.median_filter). As it was suggested in #87, this method has undefined behaviour for `NaN` values, giving inconsistent results.

There are two extra methods proposed at the moment:

- `_median_filter_2d`: based on a cython implementation of a 2D median filter. This method handles `NaN` values predictably, by ignoring them when creating the 2D-kernel for filtering. In the case a kernel has an even number of elements, the average value of the two middle elements is selected as the median, which is what [`numpy.nanmedian`](https://numpy.org/doc/stable/reference/generated/numpy.nanmedian.html#) does.

- `_median_filter_nd`: is an alternative for n-dimensional filtering, which is the main advantage of Scipy's median filter. This function uses `numpy.sliding_window_view` and `numpy.nanmedian` to perform the median filtering with n-dimensional kernels.  This function also ignores `NaN` values when performing the median calculation for the kernel, with the same criteria as the `_median_filter_2d` method.

Regarding execution times:
- The fastest method for 2D filtering is `_median_filter_2d`, which is marginally faster (10-15%) than `_median_filter_scipy`.
- `_median_filter_nd` is around 3-6 times slower than `_median_filter_scipy` for any dimensional arrays.

## Checklist

### DIFF
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2093,48 +2093,53 @@ cdef inline float_t _median(float_t *values, const ssize_t n) noexcept nogil:
 def _apply_2d_median_filter(
     const float_t[:, ::1] image,
     float_t[:, ::1] filtered_image,
-    const ssize_t kernel_size
+    const ssize_t kernel_size,
+    const int num_threads=1,
 ):
     """Apply a 2D median filter ignoring NaN."""
     cdef:
         ssize_t rows = image.shape[0]
         ssize_t cols = image.shape[1]
         ssize_t k = kernel_size // 2
-        ssize_t i, j, di, dj
-        ssize_t ki, kj
-        ssize_t valid_count
+        ssize_t i, j, di, dj, ki, kj, valid_count
         float_t element
-        float_t *kernel = <float_t *>malloc(
-            kernel_size * kernel_size * sizeof(float_t)
-        )
+        float_t *kernel
 
     if kernel_size <= 0:
         raise ValueError("kernel_size must be greater than 0")
-    if kernel == NULL:
-        raise MemoryError("Unable to allocate memory for kernel")
 
-    with nogil:
-        for i in range(rows):
+    with nogil, parallel(num_threads=num_threads):
+
+        kernel = <float_t *> malloc(
+            kernel_size * kernel_size * sizeof(float_t)
+        )
+        if kernel == NULL:
+            with gil:
+                raise MemoryError("Unable to allocate memory for kernel")
+
+        for i in prange(rows):
             for j in range(cols):
-                if not isnan(image[i, j]):
-                    valid_count = 0
-                    for di in range(kernel_size):
-                        ki = i - k + di
-                        if ki < 0:
-                            ki = 0
-                        elif ki >= rows:
-                            ki = rows - 1
-                        for dj in range(kernel_size):
-                            kj = j - k + dj
-                            if kj < 0:
-                                kj = 0
-                            elif kj >= cols:
-                                kj = cols - 1
-                            element = image[ki, kj]
-                            if not isnan(element):
-                                kernel[valid_count] = element
-                                valid_count += 1
-                    if valid_count > 0:
-                        filtered_image[i, j] = _median(kernel, valid_count)
+                if isnan(image[i, j]):
+                    filtered_image[i, j] = <float_t> NAN
+                    continue
+                valid_count = 0
+                for di in range(kernel_size):
+                    ki = i - k + di
+                    if ki < 0:
+                        ki = 0
+                    elif ki >= rows:
+                        ki = rows - 1
+                    for dj in range(kernel_size):
+                        kj = j - k + dj
+                        if kj < 0:
+                            kj = 0
+                        elif kj >= cols:
+                            kj = cols - 1
+                        element = image[ki, kj]
+                        if not isnan(element):
+                            kernel[valid_count] = element
+                            valid_count = valid_count + 1
+                if valid_count > 0:
+                    filtered_image[i, j] = _median(kernel, valid_count)
 
-    free(kernel)
+        free(kernel)
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -1866,7 +1866,7 @@ cdef float_t _quickselect(
     float_t *arr,
     ssize_t left,
     ssize_t right,
-    ssize_t k
+    const ssize_t k
 ) noexcept nogil:
     """Quickselect algorithm to find the k-th smallest element."""
     cdef:
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2086,12 +2086,12 @@ cdef float_t _median(const float_t *values, const ssize_t n) noexcept nogil:
             (_quickselect(values, 0, n - 1, n // 2 - 1) +
              _quickselect(values, 0, n - 1, n // 2)) / 2
         )
-    
+
     return _quickselect(values, 0, n - 1, n // 2)
 
 
 def _apply_2d_median_filter(
-    const float_t[:, :] image, 
+    const float_t[:, :] image,
     float_t[:, :] filtered_image,
     const ssize_t kernel_size
 ):
@@ -2103,7 +2103,7 @@ def _apply_2d_median_filter(
         int rows = image.shape[0]
         int cols = image.shape[1]
         int k = kernel_size // 2
-        int i, j, di, dj, iter
+        int i, j, di, dj
         int ki, kj
         int valid_count
         float_t element
@@ -2113,7 +2113,7 @@ def _apply_2d_median_filter(
 
     if kernel == NULL:
         raise MemoryError("Unable to allocate memory for kernel")
-    
+
     with nogil:
         for i in range(rows):
             for j in range(cols):
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -1940,7 +1940,5 @@ def _median_filter_2d(float_t[:, :] image, float_t[:, :] filtered_image,
                             valid_count += 1
                 if valid_count > 0:
                     filtered_image[i, j] = _median(kernel, valid_count)
-            else:
-                print(image[i, j])
 
     free(kernel)
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -61,6 +61,8 @@ ctypedef fused signal_t:
     float
     double
 
+from libc.stdlib cimport free, malloc
+
 
 def _phasor_from_signal(
     float_t[:, :, ::1] phasor,
@@ -1855,3 +1857,92 @@ cdef float_t _anscombe_inverse_approx(
         + 0.7654655446197431 / (x * x * x)  # 5/8 * sqrt(3/2)
         - 0.125  # 1/8
     )
+
+
+###############################################################################
+# Filtering functions
+
+cdef float_t _quickselect(float_t *arr, int left, int right, int k) nogil:
+    """Quickselect algorithm to find the k-th smallest element."""
+    cdef int i, j, pivotIndex, pivotNewIndex
+    cdef float_t pivotValue, temp
+
+    while left <= right:
+        pivotIndex = left + (right - left) // 2
+        pivotValue = arr[pivotIndex]
+        temp = arr[pivotIndex]
+        arr[pivotIndex] = arr[right]
+        arr[right] = temp
+        pivotNewIndex = left
+        for i in range(left, right):
+            if arr[i] < pivotValue:
+                temp = arr[i]
+                arr[i] = arr[pivotNewIndex]
+                arr[pivotNewIndex] = temp
+                pivotNewIndex += 1
+        temp = arr[right]
+        arr[right] = arr[pivotNewIndex]
+        arr[pivotNewIndex] = temp
+
+        if pivotNewIndex == k:
+            return arr[k]
+        elif pivotNewIndex < k:
+            left = pivotNewIndex + 1
+        else:
+            right = pivotNewIndex - 1
+
+    return arr[k]
+
+
+cdef float_t _median(float_t *values, int n) nogil:
+    """Calculate the median of an array of values."""
+    if n % 2 == 0:
+        return (
+            (_quickselect(values, 0, n - 1, n // 2 - 1) +
+             _quickselect(values, 0, n - 1, n // 2)) / 2
+        )
+    else:
+        return _quickselect(values, 0, n - 1, n // 2)
+
+
+def _median_filter_2d(float_t[:, :] image, float_t[:, :] filtered_image,
+                      const ssize_t kernel_size):
+    """Apply a 2D median filter ignoring NaN."""
+    cdef:
+        int rows = image.shape[0]
+        int cols = image.shape[1]
+        int k = kernel_size // 2
+        int i, j, di, dj, iter
+        int ki, kj
+        int valid_count
+        float_t *kernel = <float_t *>malloc(
+            kernel_size * kernel_size * sizeof(float_t)
+        )
+
+    for i in range(rows):
+        for j in range(cols):
+            if isnan(image[i, j]):
+                filtered_image[i, j] = <float_t>NAN
+                continue
+            valid_count = 0
+            for di in range(kernel_size):
+                for dj in range(kernel_size):
+                    ki = i - k + di
+                    kj = j - k + dj
+                    if ki < 0:
+                        ki = 0
+                    elif ki >= rows:
+                        ki = rows - 1
+                    if kj < 0:
+                        kj = 0
+                    elif kj >= cols:
+                        kj = cols - 1
+                    if not isnan(image[ki, kj]):
+                        kernel[valid_count] = image[ki, kj]
+                        valid_count += 1
+            if valid_count > 0:
+                filtered_image[i, j] = _median(kernel, valid_count)
+            else:
+                filtered_image[i, j] = <float_t>NAN
+
+    free(kernel)
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -1921,28 +1921,26 @@ def _median_filter_2d(float_t[:, :] image, float_t[:, :] filtered_image,
 
     for i in range(rows):
         for j in range(cols):
-            if isnan(image[i, j]):
-                filtered_image[i, j] = <float_t>NAN
-                continue
-            valid_count = 0
-            for di in range(kernel_size):
-                for dj in range(kernel_size):
-                    ki = i - k + di
-                    kj = j - k + dj
-                    if ki < 0:
-                        ki = 0
-                    elif ki >= rows:
-                        ki = rows - 1
-                    if kj < 0:
-                        kj = 0
-                    elif kj >= cols:
-                        kj = cols - 1
-                    if not isnan(image[ki, kj]):
-                        kernel[valid_count] = image[ki, kj]
-                        valid_count += 1
-            if valid_count > 0:
-                filtered_image[i, j] = _median(kernel, valid_count)
+            if not isnan(image[i, j]):
+                valid_count = 0
+                for di in range(kernel_size):
+                    for dj in range(kernel_size):
+                        ki = i - k + di
+                        kj = j - k + dj
+                        if ki < 0:
+                            ki = 0
+                        elif ki >= rows:
+                            ki = rows - 1
+                        if kj < 0:
+                            kj = 0
+                        elif kj >= cols:
+                            kj = cols - 1
+                        if not isnan(image[ki, kj]):
+                            kernel[valid_count] = image[ki, kj]
+                            valid_count += 1
+                if valid_count > 0:
+                    filtered_image[i, j] = _median(kernel, valid_count)
             else:
-                filtered_image[i, j] = <float_t>NAN
+                print(image[i, j])
 
     free(kernel)
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2091,7 +2091,7 @@ cdef inline float_t _median(float_t *values, const ssize_t n) noexcept nogil:
 
 
 def _apply_2d_median_filter(
-    const float_t[:, ::1] image,
+    float_t[:, :] image,
     float_t[:, ::1] filtered_image,
     const ssize_t kernel_size,
     const int num_threads=1,
@@ -2142,3 +2142,7 @@ def _apply_2d_median_filter(
                 filtered_image[i, j] = _median(kernel, valid_count)
 
         free(kernel)
+
+        for i in prange(rows):
+            for j in range(cols):
+                image[i, j] = filtered_image[i, j]
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2139,7 +2139,6 @@ def _apply_2d_median_filter(
                         if not isnan(element):
                             kernel[valid_count] = element
                             valid_count = valid_count + 1
-                if valid_count > 0:
-                    filtered_image[i, j] = _median(kernel, valid_count)
+                filtered_image[i, j] = _median(kernel, valid_count)
 
         free(kernel)
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2104,7 +2104,9 @@ def _apply_2d_median_filter(
         ssize_t ki, kj
         ssize_t valid_count
         float_t element
-        float_t *kernel
+        float_t *kernel = <float_t *>malloc(
+            kernel_size * kernel_size * sizeof(float_t)
+        )
 
     if kernel_size <= 0:
         raise ValueError("kernel_size must be greater than 0")
@@ -2113,30 +2115,26 @@ def _apply_2d_median_filter(
 
     with nogil:
         for i in range(rows):
-            kernel = <float_t *>malloc(kernel_size * kernel_size * sizeof(float_t))
-            if kernel == NULL:
-                raise MemoryError("Unable to allocate memory for kernel")
-            try:
-                for j in range(cols):
-                    if not isnan(image[i, j]):
-                        valid_count = 0
-                        for di in range(kernel_size):
-                            ki = i - k + di
-                            if ki < 0:
-                                ki = 0
-                            elif ki >= rows:
-                                ki = rows - 1
-                            for dj in range(kernel_size):
-                                kj = j - k + dj
-                                if kj < 0:
-                                    kj = 0
-                                elif kj >= cols:
-                                    kj = cols - 1
-                                element = image[ki, kj]
-                                if not isnan(element):
-                                    kernel[valid_count] = element
-                                    valid_count += 1
-                        if valid_count > 0:
-                            filtered_image[i, j] = _median(kernel, valid_count)
-            finally:
-                free(kernel)
+            for j in range(cols):
+                if not isnan(image[i, j]):
+                    valid_count = 0
+                    for di in range(kernel_size):
+                        ki = i - k + di
+                        if ki < 0:
+                            ki = 0
+                        elif ki >= rows:
+                            ki = rows - 1
+                        for dj in range(kernel_size):
+                            kj = j - k + dj
+                            if kj < 0:
+                                kj = 0
+                            elif kj >= cols:
+                                kj = cols - 1
+                            element = image[ki, kj]
+                            if not isnan(element):
+                                kernel[valid_count] = element
+                                valid_count += 1
+                    if valid_count > 0:
+                        filtered_image[i, j] = _median(kernel, valid_count)
+
+    free(kernel)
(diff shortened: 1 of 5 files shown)


bruno-pannunzio:
Hi @cgohlke! I am working on improving the median filter for consistency when handling `NaN` values. As expected, it seems it is a trade-off between consistency, performance and adaptability for multiple dimensions. Speaking with Leonel, a couple of points were raised regarding filtering phasor coordinates:
- He suggests prioritizing consistency before the other aspects
- Phasor coordinates are usually filtered in the 2D space, at least for microscopy images. This is related to the resolution along different axes when acquiring images.

Considering this, I testes multiple approaches to replace the median filtering from scipy, since we know is not consistent at all with `NaN` as I also testes in various conditions and is not precise and predictable at all. I implemented two different methods for filtering, one based on a cython function with only works with 2D arrays and performs very similar to Scipy's. I also implemented another method that uses `numpy.nanmedian` which is very consistent with `NaN` values also. The advantage of this method is that it can perform filtering in n-dimensions but the drawback is that is much slower, and filtering is already a very costly function. 

I left the three proposed methods to discuss in this PR, but I believe the best approach would be the following:
- Remove the `scipy_median` method, which would also remove Scipy as a dependency (I think).
- Use only one method `median` that uses the cython implementation is filtering in 2D and uses the `multidim_median` implementation if filtering in higher dimensions.

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
         + 0.7654655446197431 / (x * x * x)  # 5/8 * sqrt(3/2)
         - 0.125  # 1/8
     )
+
+
+###############################################################################
+# Filtering functions
+
+cdef float_t _quickselect(float_t *arr, int left, int right, int k) nogil:
```
cgohlke, line 1865: - All `int` types for indices and sizes should be replaced by `ssize_t` type in order to be able to process > 2 GB images.
- `nogil` -> `noexcept nogil`
- Use `const` qualifiers if possible.

`cdef float_t _quickselect(const float_t *arr, ssize_t left, ssize_t right, ssize_t k) noexcept nogil:`

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+    """Quickselect algorithm to find the k-th smallest element."""
+    cdef int i, j, pivotIndex, pivotNewIndex
+    cdef float_t pivotValue, temp
```
cgohlke, line 1868: Follow Cython/Python code styles:
```
    cdef:
        ssize_t i, j, pivot_index, pivot_new_index
        float_t pivot_value, temp
```

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+
+
+cdef float_t _median(float_t *values, int n) nogil:
+    """Calculate the median of an array of values."""
+    if n % 2 == 0:
+        return (
+            (_quickselect(values, 0, n - 1, n // 2 - 1) +
+             _quickselect(values, 0, n - 1, n // 2)) / 2
+        )
+    else:
```
cgohlke, line 1904: `else` not needed

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+            return arr[k]
+        elif pivotNewIndex < k:
+            left = pivotNewIndex + 1
+        else:
+            right = pivotNewIndex - 1
+
+    return arr[k]
+
+
+cdef float_t _median(float_t *values, int n) nogil:
```
cgohlke, line 1897: `cdef float_t _median(const float_t *values, const ssize_t n) noexcept nogil:`

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+        int cols = image.shape[1]
+        int k = kernel_size // 2
+        int i, j, di, dj, iter
+        int ki, kj
+        int valid_count
+        float_t *kernel = <float_t *>malloc(
+            kernel_size * kernel_size * sizeof(float_t)
+        )
+
+    for i in range(rows):
```
cgohlke, line 1922: Release the GIL before: `with nogil:`

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+                      const ssize_t kernel_size):
+    """Apply a 2D median filter ignoring NaN."""
+    cdef:
+        int rows = image.shape[0]
+        int cols = image.shape[1]
+        int k = kernel_size // 2
+        int i, j, di, dj, iter
+        int ki, kj
+        int valid_count
+        float_t *kernel = <float_t *>malloc(
```
cgohlke, line 1918: Must check for failed `malloc` (`kernel == NULL`) and raise `MemoryError` if needed.

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+        return _quickselect(values, 0, n - 1, n // 2)
+
+
+def _median_filter_2d(float_t[:, :] image, float_t[:, :] filtered_image,
+                      const ssize_t kernel_size):
+    """Apply a 2D median filter ignoring NaN."""
+    cdef:
+        int rows = image.shape[0]
+        int cols = image.shape[1]
+        int k = kernel_size // 2
```
cgohlke, line 1914: Verify that `kernel_size` > 0

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+        return (
+            (_quickselect(values, 0, n - 1, n // 2 - 1) +
+             _quickselect(values, 0, n - 1, n // 2)) / 2
+        )
+    else:
+        return _quickselect(values, 0, n - 1, n // 2)
+
+
+def _median_filter_2d(float_t[:, :] image, float_t[:, :] filtered_image,
+                      const ssize_t kernel_size):
```
cgohlke, line 1909: Use black format:

```
def _median_filter_2d(
    float_t[:, :] image, 
    float_t[:, :] filtered_image,
    const ssize_t kernel_size
):
```

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+            if not isnan(image[i, j]):
+                valid_count = 0
+                for di in range(kernel_size):
+                    for dj in range(kernel_size):
+                        ki = i - k + di
+                        kj = j - k + dj
+                        if ki < 0:
+                            ki = 0
+                        elif ki >= rows:
+                            ki = rows - 1
```
cgohlke, line 1933: `ki` handling can be moved one loop-level up.

src/phasorpy/phasor.py
```diff
@@ -3070,13 +3092,157 @@ def _median(
     return numpy.nanmedian(real, **kwargs), numpy.nanmedian(imag, **kwargs)
 
 
-def _median_filter(
+def _2d_median_filter(
```
cgohlke, line 3095: Better `_median_filter_2d`?

src/phasorpy/phasor.py
```diff
@@ -3070,13 +3092,157 @@ def _median(
...
+            _median_filter_2d(real[full_index], real_slice, size)
+            _median_filter_2d(imag[full_index], imag_slice, size)
+
+            real[full_index] = real_slice
+            imag[full_index] = imag_slice
+
+    return numpy.asarray(real), numpy.asarray(imag)
+
+
+def _multidim_median_filter(
```
cgohlke, line 3165: Better `_median_filter_nd`?

src/phasorpy/phasor.py
```diff
@@ -3070,13 +3092,157 @@ def _median(
...
+        imag = numpy.nanmedian(
+            windows_imag, axis=tuple(range(-len(kernel_shape), 0))
+        )
+        real = numpy.where(nan_mask, numpy.nan, numpy.asarray(real))
+        imag = numpy.where(nan_mask, numpy.nan, numpy.asarray(imag))
+
+    return real, imag
+
+
+def _scipy_median_filter(
```
cgohlke, line 3236: Better `_median_filter_scipy`?

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+                        if kj < 0:
+                            kj = 0
+                        elif kj >= cols:
+                            kj = cols - 1
+                        if not isnan(image[ki, kj]):
```
cgohlke, line 1938: Probably worth assigning `image[ki, kj]` to a temporary variable instead of looking it up twice in the innermost loop.

src/phasorpy/phasor.py
```diff
@@ -2672,10 +2673,14 @@ def phasor_filter(
     imag: ArrayLike,
     /,
     *,
-    method: Literal['median'] = 'median',
+    method: Literal[
+        '2d_median', 'multidim_median', 'scipy_median'
```
cgohlke, line 2677: Better `'median_2d', 'median_nd', 'median_scipy'`?

src/phasorpy/_phasorpy.pyx
```diff
@@ -1855,3 +1857,88 @@ cdef float_t _anscombe_inverse_approx(
...
+                arr[i] = arr[pivotNewIndex]
+                arr[pivotNewIndex] = temp
+                pivotNewIndex += 1
+        temp = arr[right]
+        arr[right] = arr[pivotNewIndex]
+        arr[pivotNewIndex] = temp
+
+        if pivotNewIndex == k:
+            return arr[k]
+        elif pivotNewIndex < k:
```
cgohlke, line 1889: `elif` -> `if`

src/phasorpy/phasor.py
```diff
@@ -2672,10 +2673,14 @@ def phasor_filter(
...
+    ] = '2d_median',
     repeat: int = 1,
+    size: int = 3,
+    skip_axis: int | None = None,
     **kwargs: Any,
-) -> tuple[NDArray[Any], NDArray[Any]]:
+) -> tuple[NDArray[numpy.float64], NDArray[numpy.float64]]:
```
cgohlke, line 2683: Why? The Cython functions work with float32 and float64, no?

cgohlke:
Hi @bruno-pannunzio , I haven't tested/run this yet but left some comments.

> Phasor coordinates are usually filtered in the 2D space, at least for microscopy images. This is related to the resolution along different axes when acquiring images.

Sure. As a library, PhasorPy should still be generic if reasonably possible.

src/phasorpy/_phasorpy.pyx
bruno-pannunzio: I tried using the `const` qualifier with `*arr` but it doesn;t seem to be compatible with `nogil`

src/phasorpy/phasor.py
```diff
@@ -3070,13 +3092,157 @@ def _median(
...
+    real = numpy.asarray(real, dtype=float).copy()
+    imag = numpy.asarray(imag, dtype=float).copy()
+
+    if axes is None:
+        axes = range(real.ndim)
+    if len(axes) != 2:
+        raise ValueError(
+            "Exactly 2 axes for filtering are required by this method."
+            f"Got {len(axes)} axes: {axes}"
+        )
```
bruno-pannunzio, line 3136: @cgohlke do you think that instead of raising this error it dispatches to `_median_filter_nd`? That way we also reduce the number of methods to maybe just `median` and `median_scipy`, in case we also want to keep scipy's?

bruno-pannunzio:
Thanks @cgohlke for all the comments! I am trying to take them all into account since I think they all make a lot of sense and improve the code.

> Sure. As a library, PhasorPy should still be generic if reasonably possible.

I totally agree with you! But discussing with Leonel we reached the conclusion that contemplating every possibility with this median filter can take a while and it was not one of the priorities of the project. We can discuss this next meeting, but at least giving the option to still filter n-dimensions leaves the door open to uses other than microscopy, although it can be optimized for performance.

I will try to add tests for this implementations to finish the PR.

src/phasorpy/_phasorpy.pyx
cgohlke: I'll look into `const float_t *arr`. Maybe use memoryviews instead.

`k` seems const too?

bruno-pannunzio: You are right, `k` is also `const`. 

I realized that `*arr` is not `const` since the algorithm changes the order of the elements to sort them. For example:

`arr[pivot_new_index] = temp`

src/phasorpy/phasor.py
cgohlke: Yes.

bruno-pannunzio:
Hi @cgohlke! I think this is ready for review considering what we discussed in this last week. Let me know if you think something can be improved.

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+###############################################################################
+# Filtering functions
+
+cdef float_t _quickselect(
+    float_t *arr,
+    ssize_t left,
+    ssize_t right,
+    const ssize_t k
+) noexcept nogil:
+    """Quickselect algorithm to find the k-th smallest element."""
```
cgohlke, line 2050: -> `"""Return k-th smallest array value using Quickselect algorithm."""`

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+
+    return arr[k]
+
+
+cdef float_t _median(const float_t *values, const ssize_t n) noexcept nogil:
+    """Calculate the median of an array of values."""
+    if n % 2 == 0:
+        return (
+            (_quickselect(values, 0, n - 1, n // 2 - 1) +
+             _quickselect(values, 0, n - 1, n // 2)) / 2
```
cgohlke, line 2087: Use `... / <float_t> 2.0`

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+            return arr[k]
+        if pivot_new_index < k:
+            left = pivot_new_index + 1
+        else:
+            right = pivot_new_index - 1
+
+    return arr[k]
+
+
+cdef float_t _median(const float_t *values, const ssize_t n) noexcept nogil:
```
cgohlke, line 2082: -> `cdef inline float_t _median(float_t *values, const ssize_t n) noexcept nogil:`

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+    """Calculate the median of an array of values."""
```
cgohlke, line 2083: -> `"""Return median of array of values."""`

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+            (_quickselect(values, 0, n - 1, n // 2 - 1) +
+             _quickselect(values, 0, n - 1, n // 2)) / 2
+        )
+
+    return _quickselect(values, 0, n - 1, n // 2)
+
+
+def _apply_2d_median_filter(
+    const float_t[:, :] image,
+    float_t[:, :] filtered_image,
```
cgohlke, line 2095: As far as I can see, you always pass C contiguous array copies to this function. Then this should be faster:
```
    const float_t[:, ::1] image,
    float_t[:, ::1] filtered_image,  # copy of image
```

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+    const ssize_t kernel_size
+):
+    """Apply a 2D median filter ignoring NaN."""
+    if kernel_size <= 0:
+        raise ValueError("kernel_size must be greater than 0")
+
+    cdef:
```
cgohlke, line 2102: Move `cdef` to top of function.

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+        float_t element
+        float_t *kernel = <float_t *>malloc(
+            kernel_size * kernel_size * sizeof(float_t)
+        )
+
+    if kernel == NULL:
+        raise MemoryError("Unable to allocate memory for kernel")
+
+    with nogil:
+        for i in range(rows):
```
cgohlke, line 2118: This could be parallelized with OpenMP. Maybe later?

src/phasorpy/_phasorpy.pyx
```diff
@@ -2034,3 +2036,106 @@ def _signal_denoise_vector(
...
+    if kernel_size <= 0:
+        raise ValueError("kernel_size must be greater than 0")
+
+    cdef:
+        int rows = image.shape[0]
+        int cols = image.shape[1]
+        int k = kernel_size // 2
+        int i, j, di, dj
+        int ki, kj
+        int valid_count
```
cgohlke, line 2108: `int` -> `ssize_t`

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3099,152 @@ def _median(
...
+) -> tuple[NDArray[Any], NDArray[Any]]:
+    """Return the phasor coordinates after applying a 2D median filter.
+
+    Parameters
+    ----------
+    real : ndarray
+        Real components of the phasor coordinates.
+    imag : ndarray
+        Imaginary components of the phasor coordinates.
+    axes : int or sequence of int
```
cgohlke, line 3119: Description does not match typing.

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3099,152 @@ def _median(
...
+        Axes along which to apply the filter. By default, filtering is
+        applied to all axes.
+    repeat : int, optional
+        Number of times to apply filter. The default is 1.
+    size : int or tuple of int, optional
```
cgohlke, line 3124: Description does not match typing.

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3099,152 @@ def _median(
...
+    real : numpy.ndarray
+        Real components of the phasor coordinates.
+    imag : numpy.ndarray
+        Imaginary components of the phasor coordinates.
+    axes : int or sequence of int
+        Axes along which to apply the filter. By default, filtering is
+        applied to all axes.
+    repeat : int, optional
+        Number of times to apply filter. The default is 1.
+    size : int or tuple of int, optional
```
cgohlke, line 3195: Description does not match typing.

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3099,152 @@ def _median(
...
+) -> tuple[NDArray[Any], NDArray[Any]]:
+    """Return the phasor coordinates after applying a median filter.
+
+    Parameters
+    ----------
+    real : numpy.ndarray
+        Real components of the phasor coordinates.
+    imag : numpy.ndarray
+        Imaginary components of the phasor coordinates.
+    axes : int or sequence of int
```
cgohlke, line 3190: Description does not match typing.

src/phasorpy/phasor.py
```diff
@@ -3087,6 +3257,9 @@ def _median_filter(
         Real components of the phasor coordinates.
     imag : ndarray
         Imaginary components of the phasor coordinates.
+    axes : int or sequence of int
+        Axes along which to apply the filter. By default, filtering is
+        applied to all axes.
     repeat : int, optional
         Number of times to apply filter. The default is 1.
     size : int or tuple of int, optional
```
cgohlke, line 3265: Description does not match typing.

src/phasorpy/phasor.py
```diff
@@ -3087,6 +3257,9 @@ def _median_filter(
         Real components of the phasor coordinates.
     imag : ndarray
         Imaginary components of the phasor coordinates.
+    axes : int or sequence of int
```
cgohlke, line 3260: Description does not match typing.

src/phasorpy/phasor.py
bruno-pannunzio: Thanks Christoph, I made some last minute changes to avoid checking some variables and forgot to change the description.

src/phasorpy/_phasorpy.pyx
bruno-pannunzio: You are right, sorry I forgot to change this.

src/phasorpy/_phasorpy.pyx
bruno-pannunzio: Thanks Christoph for this suggestion! This helped a little bit and now the Cython implementation is around 10-15% faster than Scipy's for 2D filtering.

bruno-pannunzio:
Thanks @cgohlke for all your comments! They all helped improve the description of variables and functions and also helped to improve a little bit the performance of the 2D filter. With all the changes now the Cython implementation of the 2D median filter is approx. 10-15% faster than Scipy's.

> This could be parallelized with OpenMP. Maybe later?

This is a good idea. I tried something to see if it was easy, but it seems it will take some time because we have to make sure in each parallelization that the kernel memory is allocated, at least with what I tried. So maybe we can open another issue to propose different optimizations for median filtering in 2D, and also propose some future Cython implementation for n-dimension median filtering.

cgohlke:
Here's a parallelized version. Also added a note that "filtered_image must be a copy of image" (is that correct?), which makes it easier to understand the implementation.

```Cython
def _apply_2d_median_filter(
    const float_t[:, ::1] image,
    float_t[:, ::1] filtered_image,
    const ssize_t kernel_size,
    const int num_threads=1,
):
    """Apply a 2D median filter ignoring NaN.

    filtered_image must be a copy of image.

    """
    cdef:
        ssize_t rows = image.shape[0]
        ssize_t cols = image.shape[1]
        ssize_t k = kernel_size // 2
        ssize_t i, j, di, dj, ki, kj, valid_count
        float_t element
        float_t *kernel

    if kernel_size <= 0:
        raise ValueError("kernel_size must be greater than 0")

    with nogil, parallel(num_threads=num_threads):

        kernel = <float_t *> malloc(
            kernel_size * kernel_size * sizeof(float_t)
        )
        if kernel == NULL:
            with gil:
                raise MemoryError("Unable to allocate memory for kernel")

        for i in prange(rows):
            for j in range(cols):
                if isnan(image[i, j]):
                    continue
                valid_count = 0
                for di in range(kernel_size):
                    ki = i - k + di
                    if ki < 0:
                        ki = 0
                    elif ki >= rows:
                        ki = rows - 1
                    for dj in range(kernel_size):
                        kj = j - k + dj
                        if kj < 0:
                            kj = 0
                        elif kj >= cols:
                            kj = cols - 1
                        element = image[ki, kj]
                        if not isnan(element):
                            kernel[valid_count] = element
                            valid_count = valid_count + 1
                if valid_count > 0:
                    filtered_image[i, j] = _median(kernel, valid_count)

        free(kernel)
```

bruno-pannunzio:
> Here's a parallelized version. 

Thanks @cgohlke! I believe this works. It's very similar to what I first thought of implementing but was getting an error when trying to raise an error without GIL, but didn't think of acquiring GIL again just to raise the error. I made the corresponding modifications to `phasor_filter` in order to be able to pass `num_threads` as parameter.

> Also added a note that "filtered_image must be a copy of image" (is that correct?)

At first I thought so, but I think that actually the filtered image doesn't necessary need to be a copy of image for this to work, it just has to be of same shape. The only aspect I think needed to change for this to work without needing to be a copy was the case when the `image[i, j]`  was `NaN`. Previously that needed to be the same in `filtered_image` because that value was omitted from replacement in `filtered_image`, but now I think this is covered. I replaced the copy of `real` and `imag` in `phasor_filter` and this works, so I think the copy is not needed.

cgohlke:
> Previously that needed to be the same in filtered_image because that value was omitted from replacement in filtered_image, but now I think this is covered

There is still the case where `valid_count=0` and `filtered_image` is not initialized, no?

cgohlke:
> filtered image doesn't necessary need to be a copy of image

In that case, you don't need to make a copy of the image in the Python code, but use `numpy.empty_like`, no?

bruno-pannunzio:
@cgohlke I made some modifications that I think improved a little bit the `_median_filter_nd` method.

I realized there isn't any test for checking that both median methods return the same output as the `median_scipy` method if there are any `NaN` values. I will add this test and then I think it is ready for review again.

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3110,157 @@ def _median(
...
+        for index in numpy.ndindex(
+            *[real.shape[ax] for ax in range(real.ndim) if ax not in axes]
+        ):
+            index_list: list[int | slice] = list(index)
+            for ax in axes:
+                index_list = index_list[:ax] + [slice(None)] + index_list[ax:]
+            full_index = tuple(index_list)
+
+            real_slice = real[full_index]
+            imag_slice = imag[full_index]
```
cgohlke, line 3178: Maybe I missed something before, but it looks like `real_slice` and `imag_slice` are not really guaranteed to be C contiguous. So the Cython type would have to be reverted to `const float_t[:, :] image`. Do the tests cover non-contiguous axes and slices?

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3110,157 @@ def _median(
...
+
+            _apply_2d_median_filter(
+                real_slice, filtered_real_slice, size, num_threads
+            )
+            _apply_2d_median_filter(
+                imag_slice, filtered_imag_slice, size, num_threads
+            )
+
+            real[full_index] = filtered_real_slice
+            imag[full_index] = filtered_imag_slice
```
cgohlke, line 3191: No need to allocate two filtered slices or index again on copy:
```Python
            filtered_slice = numpy.empty_like(real_slice)

            _apply_2d_median_filter(
                real_slice, filtered_slice, size, num_threads
            )
            numpy.copyto(real_slice, filtered_slice)

            _apply_2d_median_filter(
                imag_slice, filtered_slice, size, num_threads
            )
            numpy.copyto(imag_slice, filtered_slice)
```

The `numpy.copyto()` function call could be avoided by moving the copy into the `_apply_2d_median_filter` function.

Also, the allocation of the filtered slice can probably be moved out of the loops. I don't think the slice shape or dtype changes.

src/phasorpy/phasor.py
bruno-pannunzio: You are right Christoph. In the most common use I think they are, but not necessarily for any case. Maybe it's a good idea to revert this just in case. The alternative would be to allow only contiguous but that would limit the application. I'm inclined to revert to `const float_t[:, :] image`. What do you think?

Also, I will some test cases for non-contiguous axes/slices.

src/phasorpy/phasor.py
bruno-pannunzio: Thanks for noticing this Christoph, I think it makes sense to make this changes. It didn't have that much of an impact in execution times, but copying in the `_apply_2d_median_filter` function has the advantage of using parallel processing which can further improve performance.

Also at first I though of moving the indexing also out of the loop: 

```Python
for index in numpy.ndindex(
    *[real.shape[ax] for ax in range(real.ndim) if ax not in axes]
):
    index_list: list[int | slice] = list(index)
    for ax in axes:
        index_list = index_list[:ax] + [slice(None)] + index_list[ax:]
    full_index = tuple(index_list)
```

But this can be a problem for large images since the indices should be stored in some form of sequence with all values  not in `axes`, so this can be quite large. And I don't think the performance improvement would be significant. But if you think it makes sense I can make the changes.

src/phasorpy/phasor.py
```diff
@@ -3070,11 +3110,158 @@ def _median(
...
+
+            _apply_2d_median_filter(
+                real_slice, filtered_slice, size, num_threads
+            )
+            _apply_2d_median_filter(
+                imag_slice, filtered_slice, size, num_threads
+            )
+
+            real[full_index] = real_slice
+            imag[full_index] = imag_slice
```
cgohlke, line 3192: These lines can be removed now that copy is done in Cython.

src/phasorpy/phasor.py
cgohlke: OK

src/phasorpy/phasor.py
```diff
@@ -2740,21 +2763,38 @@ def phasor_filter(
...
+    methods: dict[str, Callable[..., Any]] = {
+        'median': _median_filter_2d,
+        'median_scipy': _median_filter_scipy,
+    }
     if method not in methods:
         raise ValueError(
             f'Method not supported, supported methods are: '
             f"{', '.join(methods)}"
         )
+    if repeat == 0 or size == 1:
```
cgohlke, line 2775: Please add tests for `repeat == 0` and `size == 1`

cgohlke (approved):

bruno-pannunzio:
Thanks @cgohlke for all your comments and improvements on this PR! They all helped make the median filtering more efficient, consistent and clear in documentation. I implemented the tests that were missing and removed the extra lines of code. 

I will merge this but it is still open for further improvements, specially the n-dimensional filtering.

PR #148, opened 2024-11-23
Add spectral vector denoising

### LINKED ISSUE #142: Implement spectral vector denoising algorithm
The spectral vector denoising algorithm is described in

- Rebecca C. Harman, Ryan T. Lang, Eric M. Kercher, Paige Leven, and Bryan Q. Spring. **Denoising multiplexed microscopy images in n-dimensional spectral space**. Biomedical Optics Express Vol. 13, Issue 8, pp. 4298-4309 (2022). https://doi.org/10.1364/BOE.463979

The algorithm is based on a weighted average calculation, with weights obtained in n-dimensional Chebyshev or Fourier space. The denoised spectra are then unmixed using least squares fit.

cgohlke:
## Description

This PR implements the spectral vector denoising algorithm as suggested in #142 and adds a `spectral_vector_denoise` function to the public API. 

The algorithm is described in

- Rebecca C. Harman, Ryan T. Lang, Eric M. Kercher, Paige Leven, and Bryan Q. Spring. **Denoising multiplexed microscopy images in n-dimensional spectral space**. Biomedical Optics Express Vol. 13, Issue 8, pp. 4298-4309 (2022). https://doi.org/10.1364/BOE.463979

Items to discuss:

- Only denoising in Fourier space is implemented. Implementing Chebyshev transform seems out of scope of the PhasorPy library. However, any spectral vector can be passed to the `spectral_vector_denoise` function by the user.
-  Unmixing the denoised signal using least squares fit is not implemented. Again, it seems out of scope of the PhasorPy library.
- As [noted](https://github.com/phasorpy/phasorpy/issues/142#issuecomment-2495175560), the algorithm is very inefficient.
- This implementation adds an option to filter out low intensity spectra, which can improve runtime for sparse images. The original spectra are returned in that case. Should the filtered-out spectra instead be set to NaN in order to be able to identify them?
- This implementation has not yet been tested against the synthetic reference data provided by Rebecca Harman.

Closes #142

## Checklist

### DIFF
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -428,6 +428,7 @@ def _gaussian_signal(
 ###############################################################################
 # FRET model
 
+
 @cython.ufunc
 cdef (double, double) _phasor_from_fret_donor(
     double omega,
@@ -952,6 +953,7 @@ cdef (float_t, float_t) _phasor_divide(
 
 ###############################################################################
 # Geometry ufuncs
+
 
 @cython.ufunc
 cdef short _is_inside_range(
@@ -1559,6 +1561,7 @@ cdef (double, double, double, double) _intersection_circle_line(
         y + (-dd * dx - fabs(dy) * rdd) / dr,
     )
 
+
 ###############################################################################
 # Blend ufuncs
 
@@ -1629,6 +1632,7 @@ cdef float_t _blend_lighten(
     if isnan(b) or isnan(a):
         return a
     return <float_t> max(a, b)
+
 
 ###############################################################################
 # Threshold ufuncs
@@ -1814,6 +1818,7 @@ cdef (double, double, double) _phasor_threshold_nan(
 ###############################################################################
 # Unary ufuncs
 
+
 @cython.ufunc
 cdef float_t _anscombe(
     float_t x,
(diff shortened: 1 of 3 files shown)


bruno-pannunzio:
Hi @cgohlke! Thanks for implementing this. I am not familiarized with the applications, but seeing it's from the spectral side we can get more input from @lmalacrida and @schutyb.

> * This implementation adds an option to filter out low intensity spectra, which can improve runtime for sparse images. The original spectra are returned in that case. Should the filtered-out spectra instead be set to NaN in order to be able to identify them?

This is great considering the executions times you stated. I think it should work as the `phasor_threshold`, setting to `NaN` values not considered, but then I don't know the implications it can have in future quantifications of the image. But considering they are low intensity pixels that are not considered importan because they are not denoised also, this can help also make more consistent with other functions.

Also considering this, does the function work well with inputs with `NaN`?

cgohlke:
> I am not familiarized with the applications, but seeing it's from the spectral side

While the naming of the algorithm and the publication are from the spectral side, I don't see why this denoising method should be limited to spectral data only. 

> I think it should work as the phasor_threshold, setting to NaN values not considered

Probably that can be made an option. On the other hand, the signal input can be pre-filtered, that is, low intensities set to NaN before calling the function.

> the function work well with inputs with NaN?

Yes, both, `NaN`s in the input signal and spectral vector should work. That should probably be explicitly tested.


I'm not sure how much more work should be put into this method/implementation:

- Due to it's inefficiency, it is not practical on real world sized data without extensive multi-threading (which is not available on macOS). Probably the method could be applied to spatial windows, that is, the denoising is limited to neighboring pixels.
- The method requires careful selection of the spectral vector and sigma to avoid smoothing over significant features in the signal. Maybe Chebyshev transform or PCA projected spectral vectors work better?
- I am unable to run the reference Matlab code. It looks to me that code does not exactly implement the algorithm described in the publication.

cgohlke:
Hi @bruno-pannunzio : I added testing for NaN handling and decided to return the original spectra, not NaN, if intensities are smaller than the threshold. If users want NaNs, they can pre-filter the input signal themselves.

cgohlke:
Here's an example to try out the filter. It takes about 20-25s on my computer:
```Python
import math

import numpy
from matplotlib import pyplot
from tifffile import Timer, imshow

from phasorpy.io import read_lsm
from phasorpy.phasor import phasor_from_signal
from phasorpy.plot import plot_phasor, plot_signal_image
from phasorpy.utils import spectral_vector_denoise

signal = read_lsm(fetch('paramecium.lsm')).data
mean, real, imag = phasor_from_signal(signal, axis=0)

with Timer('spectral_vector_denoise'):
    denoised = spectral_vector_denoise(
        signal,
        axis=0,
        sigma=0.1,
        vmin=15,
        # harmonic='all',
        # harmonic=[1, 2],
        # harmonic=[1, 2, 3, 4],
        num_threads=0,
    )

mean1, real1, imag1 = phasor_from_signal(denoised, axis=0)

assert numpy.allclose(mean, mean1)

plot_signal_image(signal, axis=0, show=False, title='Signal')
plot_signal_image(denoised, axis=0, show=False, title='Denoised')

plot_phasor(real, imag, allquadrants=True, show=False, title='Signal')
plot_phasor(real1, imag1, allquadrants=True, show=False, title='Denoised')

residuals = signal - denoised
vmin = residuals.min()
vmax = residuals.max()
vmax = math.ceil(max(abs(vmin), abs(vmax)))
imshow(
    residuals,
    vmin=-vmax,
    vmax=vmax,
    show=False,
    cmap='coolwarm',
    title='Residuals',
)

index = 261, 107
index = 248, 63
pyplot.figure()
pyplot.plot(signal[:, *index], label='Signal')
pyplot.plot(denoised[:, *index], label='Denoised')
pyplot.legend()
pyplot.show()
```

cgohlke:
The largest potential for improving speed is probably utilizing the symmetry of pixel distances (i,j is the same as j,i). That would require flattening the loops and syncing threads. Other options are approximating the `exp` function and re-implementing in C/C++ or even CUDA/OpenCL. These optimizations could be considered if spectral vector denoising turns out to be popular.

bruno-pannunzio (approved):
Hi @cgohlke! I tested with the code you provided and in my M1 mac it took 2 min and 25 sec. 

> decided to return the original spectra, not NaN, if intensities are smaller than the threshold. If users want NaNs, they can pre-filter the input signal themselves.

I agree, this makes a lot of sense and makes it more versatile also since you can choose to keep the original signal.

>  These optimizations could be considered if spectral vector denoising turns out to be popular.

This is a good idea. I think we can merge this PR and open an issue to make this visible, so that people can request this if it is popular, and also if someone wants to implement this optimizations themselves.

cgohlke:
Let's merge this as is for now and reconvene once we met with Bryan.

cgohlke:
Thanks @bruno-pannunzio 

> the code you provided and in my M1 mac it took 2 min and 25 sec.

That matches the 20-25s runtime I obtained with with 8 threads on Windows.

> open an issue to make this visible, so that people can request this if it is popular, and also if someone wants to implement this optimizations themselves.

I reopened #142 and moved the outstanding items to discuss there.

PR #150, opened 2024-11-28
Improve median filter

cgohlke:
## Description

This PR improves upon #147, simplifying the implementation and boosting performance of median filtering:

- Do not call `_quickselect` twice in `_median` function for even-sized kernels.
  The functions are now merged into one `_median` function.
- Move "repeat" loop into `_apply_2d_median_filter` function nogil section.
- Separate "repeat" loops for real and imag to improve memory efficiency.
- Merge `_median_filter_2d` and `_median_filter_nd` into one `_median_filter` function.
- Rename `_apply_2d_median_filter` to `_median_filter_2d`.
- Use separate `nan_mask` for real and imag to match Cython implementation.
- ~Move median filter after thresholding in introductory tutorial. That should be better practice now that the median filter correctly handles NaN, no? Added a note that "filtered coordinates can no longer be used to reconstruct the original signal" (is there a better phrasing? Should that be added to the function docstring?)~

## Checklist

### DIFF
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -2041,72 +2041,66 @@ def _signal_denoise_vector(
 ###############################################################################
 # Filtering functions
 
-cdef float_t _quickselect(
-    float_t *arr,
-    ssize_t left,
-    ssize_t right,
-    const ssize_t k
-) noexcept nogil:
-    """Return k-th smallest array value using Quickselect algorithm."""
+
+cdef float_t _median(float_t *values, const ssize_t size) noexcept nogil:
+    """Return median of array values using Quickselect algorithm."""
     cdef:
-        ssize_t i, pivot_index, pivot_new_index
+        ssize_t i, pivot_index, pivot_index_new
+        ssize_t left = 0
+        ssize_t right = size - 1
+        ssize_t middle = size // 2
         float_t pivot_value, temp
+
+    if size % 2 == 0:
+        middle -= 1  # Quickselect sorts on right
 
     while left <= right:
         pivot_index = left + (right - left) // 2
-        pivot_value = arr[pivot_index]
-        temp = arr[pivot_index]
-        arr[pivot_index] = arr[right]
-        arr[right] = temp
-        pivot_new_index = left
+        pivot_value = values[pivot_index]
+        temp = values[pivot_index]
+        values[pivot_index] = values[right]
+        values[right] = temp
+        pivot_index_new = left
         for i in range(left, right):
-            if arr[i] < pivot_value:
-                temp = arr[i]
-                arr[i] = arr[pivot_new_index]
-                arr[pivot_new_index] = temp
-                pivot_new_index += 1
-        temp = arr[right]
-        arr[right] = arr[pivot_new_index]
-        arr[pivot_new_index] = temp
+            if values[i] < pivot_value:
+                temp = values[i]
+                values[i] = values[pivot_index_new]
+                values[pivot_index_new] = temp
+                pivot_index_new += 1
+        temp = values[right]
+        values[right] = values[pivot_index_new]
+        values[pivot_index_new] = temp
 
-        if pivot_new_index == k:
-            return arr[k]
-        if pivot_new_index < k:
-            left = pivot_new_index + 1
+        if pivot_index_new == middle:
+            if size % 2 == 0:
+                return (values[middle] + values[middle + 1]) / <float_t> 2.0
+            return values[middle]
+        if pivot_index_new < middle:
+            left = pivot_index_new + 1
         else:
-            right = pivot_new_index - 1
+            right = pivot_index_new - 1
 
-    return arr[k]
-
-
-cdef inline float_t _median(float_t *values, const ssize_t n) noexcept nogil:
-    """Return median of array of values."""
-    if n % 2 == 0:
-        return (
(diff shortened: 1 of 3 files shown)


src/phasorpy/_phasorpy.pyx
```diff
@@ -2041,72 +2041,66 @@ def _signal_denoise_vector(
...
+                return (values[middle] + values[middle + 1]) / <float_t> 2.0
+            return values[middle]
+        if pivot_index_new < middle:
+            left = pivot_index_new + 1
         else:
-            right = pivot_new_index - 1
+            right = pivot_index_new - 1
 
-    return arr[k]
+    return values[middle]  # unreachable code?
```
bruno-pannunzio, line 2083: You are right Christoph, I think this line can be deleted since in all cases, the while loop will eventually find the median and return it.

src/phasorpy/phasor.py
```diff
@@ -2682,8 +2682,8 @@ def phasor_filter(
 ) -> tuple[NDArray[Any], NDArray[Any]]:
     """Return filtered phasor coordinates.
 
-    By default, a median filter is applied to the real and imaginary
-    components of phasor coordinates once with a kernel size of 3
+    By default, a median filter is applied independently to the real and
```
bruno-pannunzio, line 2685: Adding the 'independently' is a good idea, since I have been asked in some occasions if this is the case and it wasn't clear before in the description.

src/phasorpy/phasor.py
```diff
@@ -2736,35 +2736,35 @@ def phasor_filter(
...
+    ...     [[3, 3, 3], [6, math.nan, 6], [4, 4, 4]],
     ...     size=3,
     ...     repeat=3,
     ... )
     (array([[0, 0, 0],
             [2, 2, 2],
             [2, 2, 2]]),
     array([[3, 3, 3],
-            [4, 4, 4],
+            [4, nan, 4],
```
bruno-pannunzio, line 2762: Good idea to show in the example how the median method handles the `NaN`

bruno-pannunzio:
Hi @cgohlke! Thanks for implementing this improvements for median filtering. The modifications helped improving a little bit more the execution times, since now it seems to be around 20% faster than Scipy's (compared to 10-15% faster from what is now in main). 

> * Use separate `nan_mask` for real and imag to match Cython implementation.

This is something we can discuss if it makes sense to apply the same logic as the `phasor_threshold` for propagating `NaN` across real and imag so that same values are invalid in all components of the phasor. But we can keep it like this I think and remember to discuss it next meeting.

> * Move median filter after thresholding in introductory tutorial. That should be better practice now that the median filter correctly handles NaN, no?

I think it would be a good idea to have the input of @lmalacrida on this. I remember he once said to me it's better practice to first filter and then threshold. I tried first thresholding and then filtering and the result phasor has wider spread and doesn't look as nice (even with the current implementation which is `NaN` consistent). 

This is an example result of first thresholding and then filtering:
image

And this is first filtering and then thresholding where you can see it is more compact and fewer pixels lie outside the semicircle:
image

> * Added a note that "filtered coordinates can no longer be used to reconstruct the original signal" (is there a better phrasing? Should that be added to the function docstring?)

This can be a good idea to add to the docstring as a warning that once performed the filter you can't go back to the signal. Also maybe add a note in the `phasor_to_signal` function stating the same warning?

src/phasorpy/_phasorpy.pyx
cgohlke: I'll leave it since formally the function has to return a value from all code paths.

cgohlke:
>  it's better practice to first filter and then threshold

OK, I reverted the changes to the tutorial since they are disputed.

> apply the same logic as the phasor_threshold for propagating NaN across real and imag so that same values are invalid in all components of the phasor.

This would need to be documented and tested. And it should be implemented for all code paths, not just one. The filter would no longer be independent for real and imag. The logic should also apply to the intensity (at least optionally). That would probably be better implemented in a separate function (?)

I think it is easier to implement and document if the median filter is applied independently (as it is now).

lmalacrida:
@bruno-pannunzio I agree that Filtering should be performed on the entire image array, regardless of thresholding or before it. This is how i believe was performed at SimFCS.

cgohlke:
>  I agree that Filtering should be performed on the entire image array, regardless of thresholding or before it. This is how i believe was performed at SimFCS.

And the scientific reason is?

cgohlke:
> That would probably be better implemented in a separate function (?)

That's what `phasor_threshold` already does by default.

cgohlke:
> This is how i believe was performed at SimFCS.

I skimmed through the SimFCS source code. At least in some places, a minimum intensity threshold is applied as part of calculating phasor coordinates, that is **before** applying median filter. Phasor coordinates for intensities below threshold are set to zero, which are then included in the median filter (I presume). Other code paths in SimFCS might or might not do things differently.

For comparison: PhasorPy does not apply any intensity threshold by default when calculating phasor coordinates and sets phasor coordinates with no intensity to NaN. The `phasor_threshold` function sets values below or above thresholds to NaN. NaN values are ignored in the median filter.

bruno-pannunzio (approved):
Christoph I think it makes sense what you propose here and keep the workflow as it is now. 

Lets keep it like this with the modifications proposed here, where everything is very well documented.

cgohlke:
Thanks @bruno-pannunzio . Do you think we should do a new release at this point?

Issue #99, opened 2024-08-13
Add methods for filtering


bruno-pannunzio:
This issue proposes the implementation of new methods for the `phasor_filter` function in the `phasor` module. For now, the only method implemented is the median, which is just a convenience wrapper for `scipy.ndimage.median_filter`. Other methods proposed (but not restricted to) are:

- [x] The Wavelet filter, first described for FLIM in DOI: [10.1088/2050-6120/aa72ab](https://iopscience.iop.org/article/10.1088/2050-6120/aa72ab) , and also in DOI: [10.1364/BOE.420953](https://doi.org/10.1364/BOE.420953).
  The Wavelet filter was already implemented in python in [pawFLIM](https://github.com/maurosilber/pawflim).
- [x] Future implementation of the median filter in Cython can be considered, specially for consistent handling of `NaN`.
- [ ] Although less commonly used with phasors, the mean filter and gaussian filter are also common in image processing.

cgohlke:
Hi @bruno-pannunzio : it looks like some of these filters and transforms are supposed to be applied to the un-normalized intensity and Fourier components? Should we add a `normalize` parameter to the `phasor_from_signal` function?

bruno-pannunzio:
Hi @cgohlke. Optimizing the calculations for these filters could be a valuable enhancement with being able to have the un-normalized intensity. If it doesn’t seem too complex, we could consider implementing it once higher-priority tasks are completed. Otherwise, we can defer it until it's necessary during the implementation of these methods.

cgohlke:
Anscombe transformation functions are added in #139

maurosilber:
Hi,

I'm the author of pawFLIM. Let me know if I can help somehow to include it here.

As it is explained in the README, it needs the number of photons (0-th harmonic), and the first two Fourier harmonics (or the N-th and 2N-th harmonics). It expects a complex array where the harmonics are in the first dimension: `(harmonics, *shape)`.

It uses the "un-normalized" phasors since (for TCSPC) the weighted average of phasors is just adding (binning?) the un-normalized ones before dividing by the total photon number. In math:

$$ R_i = N_i r_i $$

where

- $R$ is the Fourier coefficient,
- $N$ the photon number and
- $r$ the phasor (for the $i$-th pixel)

Then:

$$ \text{WeightedAverage}(r_1, r_2) = \frac{R_1 + R_2}{N_1 + N_2} $$

[...]

PR #155, opened 2024-12-08
Fix averaging phasor coordinates and related issues

### LINKED ISSUE #122: Improve `phasor_threshold` to handle multiple harmonics simultaneously.
Current implementation of `phasor_threshold` exploits numpy broadcasting. This is not optimal in the case where multiple harmonics are present, since `mean` and `real` and `imag` have different dimensions (`mean` doesn't have the harmonics dimension) and the product of thresholding all harmonics at once is a new dimension for `mean` for  the harmonic, which from the theoretical point of view, doesn't make sense.

Also, `NaN` values should be propagated to all harmonics and components, which is not being done currently by `phasor_threshold`

For now, processing harmonics separately is recommended to avoid this problems, until a solution is implemented.

cgohlke:
## Description

This PR proposes several new features and breaking changes to fix issues raised in #99 and #122:

- Add `normalize` parameter to `phasor_from_signal` function.

  This is a trivial change but might be useful for experimenting with filter algorithms.

- Add `phasor_normalize` function.

  Use to normalize phasor coordinates returned by `phasor_from_signal(..., normalize=False)` or provided by user.

  Do we need a `phasor_unnormalize` function, maybe for internal use? It's probably a good idea that no other public function should require unnormalized coordinates. Un/re-normalization can be transparently handled internally if necessary.

- Rename `phasor_filter` function to `phasor_filter_median`:

  - Remove `method` parameter.
  - Add `use_scipy` parameter to be used instead of `method='median_scipy'`.
  - Add `mean` parameter to detect presence of harmonic axis in `real` and `imag` and make the API like other functions.
  - Remove `'axes'` from `kwargs`. It's redundant to `skip_axis`.
  - Return `mean` input unchanged to make API match other functions.
  - Remove private `_median_filter_scipy` and `_median_filter` functions. Their implementation moved into `phasor_filter_median`.

- Support multiple harmonics in `phasor_threshold`:

  - Propagate NaNs in `mean` or any `real` and `imag` harmonics to `mean` and all harmonics in `real` and `imag`.
  - Add `detect_harmonics` option to skip detecting multiple harmonics and revert to pure ufunc behavior.

- Fix `phasor_center` function:

  - Add `mean` parameter to fix averaging phasor coordinates, detect presence of harmonic axis in `real` and `imag`, and make API similar with other functions.
  - Use `phasor_threshold` for `method='mean'` to propagate NaNs among arrays to ensure the averages are calculated over the same elements. Should this be done also for `method='median'`?
  - Return filtered `mean` array, required to correctly interpret averaged phasor coordinates and make the API like other functions.
  - Change `skip_axis` refer to axes in `mean`, not `real`. The harmonic axis, if present, is automatically excluded.

- Fix `phasor_calibrate` function:

  - Add `reference_mean` parameter, which is required by the `phasor_center` function and used to automatically detect presence of harmonics.
  - Change `skip_axis` refer to axes in `reference_mean`, not `reference_real`. The harmonic axis, if present, is automatically excluded.

- Remove formal docstrings from private `_mean` and `_median` functions.

  They are repetitive, distracting, and not really needed to understand the functions.

- Adjust tests and tutorials for the above changes and features:

  Multi-harmonic support in `phasor_calibrate` and `phasor_filter_median` make the tutorials cleaner.

Closes #122.

## Checklist

### DIFF
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -971,34 +971,6 @@ cdef (float_t, float_t) _phasor_divide(
     )
 
 
-# @cython.ufunc
-# cdef (float_t, float_t) _phasor_divide_real(
-#     float_t real,
-#     float_t imag,
-#     float_t divisor,
-# ) noexcept nogil:
-#     """Return phasor coordinates divided by real number."""
-#     if divisor != 0.0:
-#         # includes isnan(divisor)
-#         return real / divisor, imag / divisor
-#     return (
-#         NAN if real == 0.0 else real * INFINITY,
-#         NAN if imag == 0.0 else imag * INFINITY
-#     )
-
-
-# @cython.ufunc
-# cdef float_t _divide(
-#     float_t dividend,
-#     float_t divisor,
-# ) noexcept nogil:
-#     """Return dividend divided by divisor."""
-#     if divisor != 0.0:
-#         # includes isnan(divisor)
-#         return dividend / divisor
-#     return NAN if dividend == 0.0 else dividend * INFINITY
-
-
 ###############################################################################
 # Geometry ufuncs
 
--- a/src/phasorpy/_phasorpy.pyx
+++ b/src/phasorpy/_phasorpy.pyx
@@ -68,7 +68,8 @@ def _phasor_from_signal(
     float_t[:, :, ::1] phasor,
     const signal_t[:, :, ::1] signal,
     const double[:, :, ::1] sincos,
-    const int num_threads
+    const bint normalize,
+    const int num_threads,
 ):
     """Return phasor coordinates from signal along middle axis.
 
@@ -97,6 +98,8 @@ def _phasor_from_signal(
         1. number samples
         2. cos and sin
 
+    normalize : bool
+        Normalize phasor coordinates.
     num_threads : int
         Number of OpenMP threads to use for parallelization.
 
@@ -145,14 +148,16 @@ def _phasor_from_signal(
                             dc = dc + sample
                             re = re + sample * sincos[h, k, 0]
                             im = im + sample * sincos[h, k, 1]
-                        if dc != 0.0:
-                            re = re / dc
-                            im = im / dc
-                            dc = dc /samples
-                        else:
-                            dc = 0.0
-                            re = NAN if re == 0.0 else re * INFINITY
-                            im = NAN if im == 0.0 else im * INFINITY
+                        if normalize:
+                            if dc != 0.0:
+                                # includes isnan(dc)
+                                re = re / dc
+                                im = im / dc
+                                dc = dc / samples
+                            else:
+                                # dc = 0.0
+                                re = NAN if re == 0.0 else re * INFINITY
+                                im = NAN if im == 0.0 else im * INFINITY
                         if h == 0:
                             mean[i, j] = <float_t> dc
                         real[h, i, j] = <float_t> re
@@ -173,14 +178,16 @@ def _phasor_from_signal(
                             dc = dc + sample
                             re = re + sample * sincos[h, k, 0]
                             im = im + sample * sincos[h, k, 1]
-                        if dc != 0.0:
-                            re = re / dc
-                            im = im / dc
-                            dc = dc /samples
-                        else:
-                            dc = 0.0
-                            re = NAN if re == 0.0 else re * INFINITY
-                            im = NAN if im == 0.0 else im * INFINITY
+                        if normalize:
+                            if dc != 0.0:
+                                # includes isnan(dc)
+                                re = re / dc
+                                im = im / dc
+                                dc = dc / samples
+                            else:
+                                # dc = 0.0
+                                re = NAN if re == 0.0 else re * INFINITY
+                                im = NAN if im == 0.0 else im * INFINITY
                         if h == 0:
                             mean[i, j] = <float_t> dc
                         real[h, i, j] = <float_t> re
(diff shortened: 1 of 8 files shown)


bruno-pannunzio:
Hi @cgohlke, sorry I wasn't able to respond before, it has been crazy this past week with the workshop.

I can review this in detail tomorrow or Friday since today is the last day of the course.

src/phasorpy/_phasorpy.pyx
```diff
@@ -924,35 +933,72 @@ cdef (float_t, float_t) _phasor_at_harmonic(
...
+#     float_t dividend,
+#     float_t divisor,
+# ) noexcept nogil:
+#     """Return dividend divided by divisor."""
+#     if divisor != 0.0:
+#         # includes isnan(divisor)
+#         return dividend / divisor
+#     return NAN if dividend == 0.0 else dividend * INFINITY
+
+
```
bruno-pannunzio, line 1001: Is it intended to keep this lines commented?

bruno-pannunzio:
Hi @cgohlke! I think all of this are great improvements to the workflow. Particularly all the tutorials are much clear now that all functions can handle harmonics. Regarding this, I also see that in the basic workflow for phasor analysis, the `plot_phasor` is the only one that doesn't support multiple harmonics and each harmonic has to be plotter separately. Do you think at some point it would make sense to modify the implementation of `plot_phasor`, or better yet the plot methods in `PhasorPlot`, to detect and plot harmonics separately? Plotting multiple harmonics in the same plot doesn't make much sense by default and it would be nice if it would support this detection of harmonics and plotting all separately. But I don't know how much work needs to be put into that.

I tested all the modifications and the results for multiple-harmonic analysis are the same as before, so everything seems to work as expected. As I said before, now the workflow is much easier not having to keep track of the harmonic axis and all related.

 > * Do we need a `phasor_unnormalize` function, maybe for internal use? It's probably a good idea that no other public function should require unnormalized coordinates. Un/re-normalization can be transparently handled internally if necessary.

I was going to propose this exactly. If it's not much trouble, it would be very nice to have this also because we are going to need this for other functions. I think it's a good idea to keep it private, at least for now.

src/phasorpy/_phasorpy.pyx
cgohlke: These lines can be removed. I needed to check that numpy divide ufuns match the Cython code we use elsewhere.

cgohlke:
> Do you think at some point it would make sense to modify the implementation of plot_phasor, or better yet the plot methods in PhasorPlot, to detect and plot harmonics separately?

`PhasorPlot` is intrinsically single harmonic. `plot_phasor` could be changed to plot a grid of single-harmonic phasor plots. Currently it can't automatically detect the presence of harmonics. We would need to add a parameter and/or pass `mean`. Anyway, that would go into a separate PR.

> If it's not much trouble, it would be very nice to have this also because we are going to need this for other functions. I think it's a good idea to keep it private, at least for now.

I'll add a private function. We'll see how useful it is. I would expect the un/normalization to be done more efficiently "inline", like in the `_mean` function, or directly in a Cython/numpy ufunc.

>> Use phasor_threshold for `method='mean'` to propagate NaNs among arrays to ensure the averages are calculated over the same elements. Should this be done also for `method='median'`?

Any thoughts on this?

bruno-pannunzio:
> * Use `phasor_threshold` for `method='mean'` to propagate NaNs among arrays to ensure the averages are calculated over the same elements. Should this be done also for `method='median'`?

I think using `phasor_threshold` for propagating NaNs is a clever way to do this. Also I think this is not computationally too heavy so it should add too much to the function. I liked how this is done. I think it can be useful also for the `median` method, but maybe I am missing something on why it should be only used with the `mean` method.

cgohlke:
> I think it can be useful also for the median method, but maybe I am missing something on why it should be only used with the mean method.

How about adding a `threshold : bool | None = None` parameter such that the user can explicitly decide if necessary (or not, for example, if the input is already thresholded)? By default, it would be enabled for `method='mean'`.

bruno-pannunzio:
> How about adding a `threshold : bool | None = None` parameter such that the user can explicitly decide

This can be a good idea to optimize for cases not needed. Although maybe there is a more intuitive name for the parameter? Because although it uses the `phasor_threshold` the idea is to propagate the NaNs, so maybe something more related to the use of the parameter? But `threshold` can work as long as the description fits.

> By default, it would be enabled for `method='mean'`

I think it's a good idea to keep it by default, but I am missing something on why it should not be default for all methods but only for mean.

cgohlke:
> Although maybe there is a more intuitive name for the parameter? 

Agreed. How about:

```
nan_safe: bool = True,

"""
nan_safe : bool, optional
    Ensure `method` is applied to same elements of input arrays.
    By default, distribute NaNs among input arrays before applying
    `method`.
"""
```

Other suggestions?

> Why it should not be default for all methods but only for mean.

Because it is not strictly necessary for median? And in another issue you argued not using thresholds before median filters (let's get back to that in Monday's meeting). Anyway, I'm in favor of applying `nan_safe=True` as a default.

bruno-pannunzio:
> nan_safe: bool = True,
> 
> """
> nan_safe : bool, optional
>     Ensure `method` is applied to same elements of input arrays.
>     By default, distribute NaNs among input arrays before applying
>     `method`.
> """

I like this very much! This is clear that is NaN related and the description matches.

> Because it is not strictly necessary for median? And in another issue you argued not using thresholds before median filters (let's get back to that in Monday's meeting).

Oh you are totally right! I think I am also inclined to using it by default. I don't know the reason behind applying the threshold before filtering, it's a good idea to discuss and try to find the reason.

cgohlke:
> If it's not much trouble, it would be very nice to have this also because we are going to need this for other functions.

I'll leave this out of this PR. Normalization is not completely reversible. If one really needs "un-normalized" coordinates, calculate them from the signal.

bruno-pannunzio (approved):
Great @cgohlke! I think this is ready considering what we discussed. 

Testing this for the wavelet filtered I encountered the same, unnormalizing after normalization is not so straight-forward. Also, I am not sure all functions work well with unnormalized data. For example, I tried the basic pipeline and I think `phasor_calibrate` is not working correctly, but I need further testing. Maybe we can consider this at least for the most commonly used functions, but we can talk about this in the meeting.

Issue #164, opened 2024-12-19
Support FLIMLABS JSON files


bruno-pannunzio:
[FLIMLABS](https://www.flimlabs.com) currently has two software for acquiring and analyzing fluorescence lifetime information:

- [Spectroscopy](https://www.flimlabs.com) for real-time photon intensity decay profile analysis, which can export files in BIN and JSON formats.
- FLIM IMAGER (not yet officially released) for FLIM image acquisition and analysis with phasor analysis capabilities. Files are exported in JSON format.

Initially, prioritize compatibility with FLIM IMAGER exported files due to the nature of the PhasorPy library and FLIM, then we can think of supporting Spectroscopy files also.

Further sample files of real images and calibration files are needed to test the implementation of the reader function for the FLIMLABS JSON files.

Below are FLIMLABS collaborators included in the development of the software, tagged for awareness of this issue. Please feel free to join the discussion and/or include any relevant details for the implementation of the reader functions of FLIMLABS JSON files.

@FLIMLABS
@AndreaNobiliFB
@aurora5bits
@MisaelSabb

cgohlke:
Hi @bruno-pannunzio : can you also share the Python file that should accompany the sample file?

The metadata in the sample file hint at 3 channels, 10 frames, 256x256 images, and 256 bins, but the data is just a nested list of 3 channels, 65536 pixels, and variable number of histogram bin indices and counts. Are the 10 frames integrated?

cgohlke:
Hi @FLIMLABS

Two suggestions:

1. Consider adding a [JSON Schema](https://json-schema.org/) for the FLIMLABS JSON image file format.
2. Consider publishing sample files and their description on [Zenodo](https://zenodo.org/) that may serve as a reference for the file format, can showcase the quality of the data from your instrumentation, and can be used for testing independent file reader implementations. Zenodo records can also be properly cited and attributed.

bruno-pannunzio:
> can you also share the Python file that should accompany the sample file?

Hi @cgohlke. I had to remove the sample file per request, since it wasn't the definitive sample file it was better not show to the public. When we have the definitive sample files I can upload the corresponding python file (with the approval of FLIMLABS) also.

FLIMLABS:
Hi @cgohlke ,

Thank you for your valuable feedback.

We will address points 1 and 2 from your suggestions and will update you once the sample files are finalized.

We also appreciate @bruno-pannunzio  for facilitating this process.

@AndreaNobiliFB
[...]

PR #165, opened 2024-12-20
WIP: Support FLIMLABS JSON image files

cgohlke:
## Description

This PR adds a `read_flimlabs_json` function to the `phasorpy.io` module to read data and metadata from FLIMLABS JSON image files. See #164.

TODO: The `read_flimlabs_json` function and docstring examples are currently not tested during CI since no public data files are available on Zenodo or GitHub.

## Checklist

### DIFF
--- a/src/phasorpy/io.py
+++ b/src/phasorpy/io.py
@@ -1063,6 +1063,7 @@ def read_flimlabs_json(
     ------
     ValueError
         File is not a FLIMLABS JSON image file.
+        `dtype` is not an unsigned integer.
 
     Examples
     --------
@@ -1085,14 +1086,14 @@ def read_flimlabs_json(
 
     with open(filename) as fh:
         try:
-            jdata = json.load(fh)
+            data = json.load(fh)
         except json.JSONDecodeError as exc:
             raise ValueError('not a FLIMLABS JSON file') from exc
 
     if (
-        'data' not in jdata
-        or 'header' not in jdata
-        or 'laser_period_ns' not in jdata['header']
+        'data' not in data
+        or 'header' not in data
+        or 'laser_period_ns' not in data['header']
     ):
         raise ValueError('not a FLIMLABS JSON file')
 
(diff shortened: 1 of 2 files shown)


bruno-pannunzio (commented):
That was fast @cgohlke! The function seems to be working correctly and I think all the possible information is extracted from the JSON file. I don't think we can use the `frames` because if I understood correctly, this JSON contains the sum of all frames (or the average?). There was an option for exporting separate frames as JSON files, but reading that can be a problem here. Maybe we can discuss with FLIMLABS if there is a possibility for exporting frames in a single JSON but mantaining the separate frames?

cgohlke:
Let's leave this PR open for now and wait for "official" sample files to be shared by FLIMLABS.

cgohlke:
Hi @bruno-pannunzio : 

Apparently FLIM LABS JSON files can contain either the signal (histogram) data or S and G, which makes the `read_flimlabs_json` return values ambiguous. We would need two functions, `signal_from_flimlabs_json` and `phasor_from_flimlabs_json`. 

At that point how about renaming all functions in `phasorpy.io` to either `signal_from_*` or `phasor_from_*`. Recently a user got (rightly) confused that `read_ifli` did not return a signal like `read_ptu` but phasor coordinates. I think renaming the functions to be less ambiguous and more in line with the rest of the library could help. 

Another issue: the `io.signal_from_*` functions would return xarray DataArrays while the `io.phasor_from_*` functions would return three numpy arrays and one metadata dict. Is this too confusing? Should we instead consider giving up on xarray and return numpy arrays and separate metadata dicts from all functions? Another option is to return xarray Datasets from `io.phasor_from_*` functions, but that is not in line with other similar named library functions, and xarray Datasets are not as transparent as DataArrays.

Btw, I originally chose to return xarray DataArrays because they are a community-accepted way to pass metadata along with array data that are mostly transparent for numpy users.

Any thoughts?

bruno-pannunzio:
Hi @cgohlke,

Those are very good points you raised. On one hand you are right that it is not consistent to always return different things from the reader functions. It is good idea to always return the same with the functions with similar name and its more intuitive. On the other hand, the 'reader' in the function name made it very obvious what was the purpose of the function. But I agree that the best option would be to rename the functions to `phasor_from_*` and `signal_from_*`, also to be consistent with `phasor_from_ometiff`.

I see the benefit of using xarray DataArrays, it is a nice way to keep the metadata and data structure very tidy and transparent. I am not familiar with xarrays Datasets, but is something we can explore. I would prefer to keep the xarray DataArrays for the moment, the seem to be working really fine with the library workflow. Since the `phasor_from_*` and `signal_from_*` will be different, I think we don't need to return the same with both, but I am also not entirely sure of this.

PR #167, opened 2024-12-27
Add pawFLIM wavelet filter for phasor coordinates

bruno-pannunzio:
## Description

This PR proposes the addition of `phasor_filter_wavelet` function in the `phasorpy.phasor` module. The implementation is based on [pawFLIM](https://github.com/maurosilber/pawflim).

The implementation of this filter was proposed and discussed in #99.

## Checklist


bruno-pannunzio:
Hi @cgohlke, I started this PR so that we can discuss some aspects on the next PhasorPy meeting if you are able to take a look at this implementation.

This is still a WIP so there are tests and documentation missing.

Based on our discussions with Mauro, I tried implementing the wavelet filter based on pawFLIM. There are some aspects that would need further discussion:
- Using pawFLIM might not be the most efficient implementation since it needs the unnormalized phasor coordinates as complex values and then internally calculates the normalized phasor coordinates again (more than once), which seems not optimized in our case since we work mostly with normalized phasor coordinates already.
- In #99 it was discussed that pawFLIM needs as input the n(th) and 2n(th) harmonic to filter the n(th) harmonic, but I believe it also filters the 2n(th) harmonic, not only the n(th). 
- From the previous point, it also needs discussion how to handle the harmonics that don't have the 2n(th) harmonic for calculation. Should they be omitted from filtering? Do we raise an error in this case?
- Maybe try another implementation of wavelet filters that don't need higher harmonics?
- The `NaN` handling was also discussed in #99. An approach I think makes sense is to convert `NaN` to 0 in the `mean`, `real` and `imag` components for filtering and then convert again to `NaN` propagating for values where the `mean` is 0. Should the `NaN` be also applied for coordinates with 0 value for `real` and `imag` when they are unnormalized (after filtering). I tested keeping the `NaN` when filtering but the results were really unsatisfactory. With just few 'levels' you loose too many pixels due to `NaN` propagation. I can show this on the next meeting.

These are some of the aspects that I believe need further discussion before moving forward with the implementation, testing, documentation, etc., for this approach to wavelet filtering.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+        `imag`.
+        If `'all'`, the first axis of `real` and `imag` contains lower
+        harmonics.
+        The default is the first harmonic (fundamental frequency).
+    skip_axis : int or sequence of int, optional
+        Axes in `mean` to exclude from filter.
+        By default, all axes except harmonics are included.
+    normalize : bool, optional
+        By default (False), expects and returns normalized phasor coordinates.
+        If True, unnormalized phasor coordinates are expected as input.
```
cgohlke, line 3048: OK for now (while still developing), but as discussed, I do not think phasorpy functions should take/return un-normalized phasor coordinates.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+    """Return wavelet-filtered phasor coordinates.
+
+    Parameters
+    ----------
+    mean : array_like
+        Intensity of phasor coordinates.
+    real : array_like
+        Real component of phasor coordinates to be filtered.
+    imag : array_like
+        Imaginary component of phasor coordinates to be filtered.
```
cgohlke, line 3026: As far as I understand, `real` and `imag` must contain at least two harmonics in the first axis? That should be documented .

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+        Controls the maximum averaging area, which has a length of 2**level.
+    harmonic : int, sequence of int, or 'all', default: 1
+        Harmonics included in `real` and `imag`.
+        If an integer, the harmonics at which `real` and `imag` were acquired
+        or calculated.
+        If a sequence, the harmonics included in the first axis of `real` and
+        `imag`.
+        If `'all'`, the first axis of `real` and `imag` contains lower
+        harmonics.
+        The default is the first harmonic (fundamental frequency).
```
cgohlke, line 3042: The default should be `'all'`, no?

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+        Axes in `mean` to exclude from filter.
+        By default, all axes except harmonics are included.
+    normalize : bool, optional
+        By default (False), expects and returns normalized phasor coordinates.
+        If True, unnormalized phasor coordinates are expected as input.
+    samples : int, optional
+        Number of samples used to calculate the phasor coordinates.
+        By default, 1.
+    nan_safe : bool, optional
+        By default, distribute NaNs among filtered arrays.
```
cgohlke, line 3053: I think this should be left out of this function.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+    sigma: float = 2.0,
+    levels: int = 1,
+    harmonic: int | Sequence[int] | Literal['all'] | str | None = None,
+    skip_axis: int | Sequence[int] | None = None,
+    normalize: bool = False,
+    samples: int = 1,
+    nan_safe: bool = True,
+) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
+    """Return wavelet-filtered phasor coordinates.
+
```
cgohlke, line 3018: Add a strong note that the function should be used with calibrated but otherwise unprocessed (never filtered or thresholded) FLIM data only.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+        Real component of phasor coordinates to be filtered.
+    imag : array_like
+        Imaginary component of phasor coordinates to be filtered.
+    sigma : float, optional
+        Significance level to test the difference between two phasors, given
+        in terms of the equivalent 1D standard deviations. sigma=2 corresponds
+        to ~95% (or 5%) significance.
+    levels : int, optional
+        Number of levels for the wavelet decomposition.
+        Controls the maximum averaging area, which has a length of 2**level.
```
cgohlke, line 3033: `2**level` -> ```:math:`2^level` ``` (I think `**` is RST syntax for bold fonts?)

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+    harmonic : int, sequence of int, or 'all', default: 1
+        Harmonics included in `real` and `imag`.
+        If an integer, the harmonics at which `real` and `imag` were acquired
+        or calculated.
```
cgohlke, line 3037: This does not apply if the method requires at least two harmonics.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+    ValueError
+        If `level` is less than 0.
+        If `samples` is less than 1.
+        If the number of harmonics in `harmonic` is less than 2.
+        If `real` and `imag` have no harmonic axis.
+        The array shapes of `mean`, `real`, and `imag` do not match.
+
+    Examples
+    --------
+
```
cgohlke, line 3075: Add reference to pawFLIM publication.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+    has_harmonic_axis = mean.ndim + 1 == real.ndim
+    if not has_harmonic_axis:
+        raise ValueError('No harmonic axis')
+    harmonic_max = real.shape[0]
+    if harmonic_max < 2:
+        raise ValueError(f'{harmonic=} < 2')
+    harmonics, _ = parse_harmonic(harmonic, harmonic_max)
+    harmonic_list = [
+        n for n in harmonics if 2 * n in range(1, harmonic_max + 1)
+    ]
```
cgohlke, line 3104: I don't think this is quite correct. The `harmonics` input parameter describes which harmonics are contained in the `real` and `imag` arrays (which may be, for example, 1 and 2, or 2 and 4, or 1 to 4), but here it is assumed that they always include 1 to `harmonic_max`.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+        The default is the first harmonic (fundamental frequency).
+    skip_axis : int or sequence of int, optional
+        Axes in `mean` to exclude from filter.
+        By default, all axes except harmonics are included.
+    normalize : bool, optional
+        By default (False), expects and returns normalized phasor coordinates.
+        If True, unnormalized phasor coordinates are expected as input.
+    samples : int, optional
+        Number of samples used to calculate the phasor coordinates.
+        By default, 1.
```
cgohlke, line 3051: The number of samples is only needed if the pawFLIM method requires the true number of photons, not relative intensity. The parameter can either be removed or should be made non-optional.

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
+
+    _, real, imag = phasor_normalize(
+        unnormalized_mean, real, imag, samples=samples
+    )
+
+    if normalize and samples > 1:
+        numpy.divide(mean, samples, out=mean)
+
+    if nan_safe:
+        return phasor_threshold(mean, real, imag, 0, open_interval=True)
```
cgohlke, line 3160: Remove

src/phasorpy/phasor.py
```diff
@@ -2998,6 +3000,168 @@ def phasor_filter_median(
...
 
+def phasor_filter_wavelet(
+    mean: ArrayLike,
+    real: ArrayLike,
+    imag: ArrayLike,
+    /,
+    *,
+    sigma: float = 2.0,
+    levels: int = 1,
+    harmonic: int | Sequence[int] | Literal['all'] | str | None = None,
```
cgohlke, line 3011: `int` can be removed if the method requires two harmonics.

src/phasorpy/phasor.py
bruno-pannunzio: You are right Christoph, the method doesn't necessarily need the number of photons, we can remove this parameter

src/phasorpy/phasor.py
bruno-pannunzio: I agree, I will then remove this parameter and require for the phasor coordinates to be normalized.

src/phasorpy/phasor.py
bruno-pannunzio: You are right @cgohlke! Thanks for pointing this out. How do you think the 'all' condition be handled? i am having some trouble thinking how to work with this. My understanding, also from the same parameter in other functions is that when `harmonic = 'all'`, it is expected to have the harmonics ordered from 1 to the max possible harmonic. Should this be more explicit in the description of the parameter? I think this concept is what 'lower harmonics' mean, but I think it would be good to say it more explicit here in the context of this function since it is very important the order of the harmonics for correct filtering.

bruno-pannunzio: Also, I think in the description of the function or in the harmonic parameter it should be detailed that for harmonics where `2 * harmonic` is not present, no filter is applied. But I am not sure how/where is the best way to describe this in the docstring.

cgohlke: >  for harmonics where 2 * harmonic is not present, no filter is applied

The filter requires h and 2h to work. It may be better to raise an error if that requirement is not met rather than silently not applying the filter.

cgohlke: Is explicitly allowing `harmonic = 'all'` really required in this function? Is the default `None` not enough? That is, by default, the phasor coordinates contain lower harmonics (`1...real.shape[0]`).

bruno-pannunzio: > The filter requires h and 2h to work. It may be better to raise an error if that requirement is not met rather than silently not applying the filter.

I agree, I don't like the idea of applying a filter only to certain harmonics and not others, so I will modify it to raise an error. This is also a good reason why searching for another implementation that doesn't need the `2 * harmonic` for all harmonics can be of great benefit here. Also making the function much simpler and broadly applicable.

> Is explicitly allowing `harmonic = 'all'` really required in this function? Is the default `None` not enough? That is, by default, the phasor coordinates contain lower harmonics (`1...real.shape[0]`).

You are right, now it doesn't make sense the `harmonic='all'` option. I will make the modifications.

bruno-pannunzio:
Hi @cgohlke, I am having some errors here but not locally in my setup. I believe it has to do with different versions of numpy. I am getting this error in `phasor_normalize`:

```
>       mean = numpy.array(
            mean_unnormalized, real.dtype, copy=None if samples == 1 else True
        )
E       ValueError: NoneType copy mode not allowed.
```

Do you know what can be happening here?

cgohlke:
CI is also testing with `numpy==1.24.4`. The `copy` parameter was changed in numpy 2. You have to find a way that works with both. See https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword and https://numpy.org/doc/1.24/reference/generated/numpy.array.html

cgohlke:
You're right. I'll fix that later today.

cgohlke:
> using phasor_normalize with samples=1, which is not covered in tests

Fixed in #186.

Anyway, `phasor_normalize` has a lot of copy overhead, which might not be necessary in your function (?)

bruno-pannunzio:
> Anyway, `phasor_normalize` has a lot of copy overhead, which might not be necessary in your function (?)

Thanks @cgohlke. You are right, I replaced the normalization by more straightforward division, which simplifies this a little bit.

I think this PR is ready for re-review. I would like to discuss also how to handle the values where `mean` is zero, or any of the phasor coordinates is `NaN`. Because with the current implementation it is replacing the invalid values with valid ones based on the filtering. This is not what `phasor_filter_median` does, which returns the `NaN` where it was present before filtering. Should we keep the same behaviour for all filtering functions?

In this case, for `phasor_filter_pawflim` I believe it would require to store a mask for the original `NaN` in phasor coordinates and apply it to the arrays after filtering. Keeping the `NaN` for filtering is a problem because it degrades a lot the result, specially as the `level` increases. I think an alternative would be to apply the `NaN` mask and propagate it with `phasor_threshold` as other functions do. What do you think is best Christoph?

cgohlke:
The implementation and API look good to me. `pawflim` should be an optional dependency. The documentation needs some fixes and clarifications. I suggest the following changes:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 7e3c6b5..574dec2 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ dependencies = [
     # "pandas>=2.0.0",
     "xarray>=2023.4.0",
     "tifffile>=2024.8.30",
-    "pawflim",
 ]
 requires-python = ">=3.10"
 classifiers = [
@@ -79,6 +78,7 @@ all = [
     "sdtfile>=2024.5.24",
     "ptufile>=2024.9.14",
     "liffile>=2025.2.6",
+    "pawflim",
 ]
 
 [tool.setuptools]
diff --git a/src/phasorpy/phasor.py b/src/phasorpy/phasor.py
index 12544ad..4e7171e 100644
--- a/src/phasorpy/phasor.py
+++ b/src/phasorpy/phasor.py
@@ -3041,34 +3041,37 @@ def phasor_filter_pawflim(
     harmonic: Sequence[int] | None = None,
     skip_axis: int | Sequence[int] | None = None,
 ) -> tuple[NDArray[Any], NDArray[Any], NDArray[Any]]:
-    """Return wavelet-filtered phasor coordinates.
+    """Return pawFLIM wavelet-filtered phasor coordinates.
 
-    This function should only be used with calibrated, unprocessed phasor
-    coordinates obtained from FLIM data. The data should not be filtered,
+    This function must only be used with calibrated, unprocessed phasor
+    coordinates obtained from FLIM data. The coordinates must not be filtered,
     thresholded, or otherwise pre-processed.
 
+    The pawFLIM wavelet filter is described in [2]_.
+
     Parameters
     ----------
     mean : array_like
         Intensity of phasor coordinates.
     real : array_like
-        Real component of phasor coordinates to be filtered. Must have at least
-        two harmonics in the first axis.
+        Real component of phasor coordinates to be filtered.
+        Must have at least two harmonics in the first axis.
     imag : array_like
-        Imaginary component of phasor coordinates to be filtered. Must have at
-        least two harmonics in the first axis.
+        Imaginary component of phasor coordinates to be filtered.
+        Must have at least two harmonics in the first axis.
     sigma : float, optional
-        Significance level to test the difference between two phasors, given
-        in terms of the equivalent 1D standard deviations. sigma=2 corresponds
-        to ~95% (or 5%) significance.
+        Significance level to test difference between two phasors.
+        Given in terms of the equivalent 1D standard deviations.
+        sigma=2 corresponds to ~95% (or 5%) significance.
     levels : int, optional
-        Number of levels for the wavelet decomposition.
-        Controls the maximum averaging area, which has a length of :math:`2^level`.
+        Number of levels for wavelet decomposition.
+        Controls the maximum averaging area, which has a length of
+        :math:`2^level`.
     harmonic : sequence of int or None, optional
-        If a sequence, the harmonics included in the first axis of `real` and
-        `imag`.
+        Harmonics included in first axis of `real` and `imag`.
         If None (default), the first axis of `real` and `imag` contains lower
         harmonics starting at and increasing by one.
+        All harmonics must have a corresponding half or double harmonic.
     skip_axis : int or sequence of int, optional
         Axes in `mean` to exclude from filter.
         By default, all axes except harmonics are included.
@@ -3095,15 +3098,14 @@ def phasor_filter_pawflim(
 
     References
     ----------
-
-    .. [1] Silberberg M & Grecco H. `pawFLIM: reducing bias and uncertainty
-    to enable lower photon count in FLIM experiments
-    <https://doi.org/10.1088/2050-6120/aa72ab>`_.
+    .. [2] Silberberg M & Grecco H. `pawFLIM: reducing bias and uncertainty
+      to enable lower photon count in FLIM experiments
+      <https://doi.org/10.1088/2050-6120/aa72ab>`_.
       *Methods Appl Fluoresc*, 5(2): 024016 (2017)
 
     Examples
     --------
-    Apply a wavelet filter with four significance levels (sigma)
+    Apply a pawFLIM wavelet filter with four significance levels (sigma)
     and three decomposition levels:
 
     >>> mean, real, imag = phasor_filter_pawflim(
@@ -3147,18 +3149,18 @@ def phasor_filter_pawflim(
 
     has_harmonic_axis = mean.ndim + 1 == real.ndim
     if not has_harmonic_axis:
-        raise ValueError('No harmonic axis')
+        raise ValueError('no harmonic axis')
     if harmonic is None:
         harmonics, _ = parse_harmonic('all', real.shape[0])
     else:
         harmonics, _ = parse_harmonic(harmonic, None)
     if len(harmonics) < 2:
         raise ValueError(
-            'At least two harmonics required, ' f'got {len(harmonics)}'
+            'at least two harmonics required, ' f'got {len(harmonics)}'
         )
     if len(harmonics) != real.shape[0]:
         raise ValueError(
-            'Number of harmonics does not match first axis of real and imag'
+            'number of harmonics does not match first axis of real and imag'
         )
 
     mean = numpy.asarray(numpy.nan_to_num(mean), dtype=float)
@@ -3173,11 +3175,11 @@ def phasor_filter_pawflim(
     _, axes = _parse_skip_axis(skip_axis, mean.ndim, True)
 
     for index in numpy.ndindex(
-        *[
+        *(
             real.shape[ax]
             for ax in range(real.ndim)
             if ax not in axes and ax != 0
-        ]
+        )
     ):
         index_list: list[int | slice] = list(index)
         for ax in axes:
@@ -3193,7 +3195,7 @@ def phasor_filter_pawflim(
                 continue
             if h * 2 not in harmonics:
                 raise ValueError(
-                    f'Harmonic {h} does not have a corresponding half '
+                    f'harmonic {h} does not have a corresponding half '
                     f'or double harmonic in {harmonics}'
                 )
             n = harmonics.index(h)
```

The tutorial could be more focused on the filters, e.g. shorten the part getting calibrated phasor coordinates and use the higher level `plot_phasor` function with default settings. I don't think `plot_signal_image` is necessary. Comparing the filtered and unfiltered images can be improved, but we don't have a nice high level function for that. How about something like this:

```Python
"""
Filter phasor coordinates
=========================

Functions for filtering phasor coordinates.

Filtering phasor coordinates improves signal quality by reducing noise while
preserving relevant features. Two of the most common methods for filtering
include the median and wavelet filtering.

"""

# %%
# Import required modules and functions:

from phasorpy.io import signal_from_imspector_tiff
from phasorpy.phasor import (
    phasor_calibrate,
    phasor_filter_median,
    phasor_filter_pawflim,
    phasor_from_signal,
    phasor_threshold,
)
from phasorpy.plot import plot_phasor

# %%
# Get calibrated phasor coordinates
# ---------------------------------
#
# Read a time-correlated single photon counting (TCSPC) histogram from a file.
# A homogeneous solution of Fluorescein (4.2 ns) was imaged as a reference:

signal = signal_from_imspector_tiff(fetch('Embryo.tif'))
frequency = signal.attrs['frequency']

reference_signal = signal_from_imspector_tiff(fetch('Fluorescein_Embryo.tif'))
assert reference_signal.attrs['frequency'] == frequency

# %%
# Calculate and calibrate phasor coordinates:

mean, real, imag = phasor_from_signal(signal, axis=0)
reference_mean, reference_real, reference_imag = phasor_from_signal(
    reference_signal, axis=0
)

real, imag = phasor_calibrate(
    real,
    imag,
    reference_mean,
    reference_real,
    reference_imag,
    frequency=frequency,
    lifetime=4.2,
)

# %%
# Plot unfiltered phasor coordinates
# ----------------------------------
#
# Plot the unfiltered, calibrated phasor coordinates after applying a
# threshold based on the mean intensity to remove background values:

plot_phasor(
    *phasor_threshold(mean, real, imag, mean_min=1)[1:],
    frequency=frequency,
    title='Unfiltered phasor coordinates',
)

# %%
# Median filter
# -------------
#
# Median filtering replaces each pixel value with the median of its
# neighboring values, reducing noise while preserving edges.
# The function :py:func:`phasorpy.phasor.phasor_filter_median` applies a
# median filter to phasor coordinates. Typically, applying a 3×3 kernel
# one to three times is sufficient to remove noise while maintaining
# important features:

mean_filtered, real_filtered, imag_filtered = phasor_filter_median(
    mean, real, imag, repeat=3, size=3
)

# %%
# Thresholds should be applied after filtering:
mean_filtered, real_filtered, imag_filtered = phasor_threshold(
    mean_filtered, real_filtered, imag_filtered, mean_min=1
)

# %%
# Plot the median-filtered and thresholded phasor coordinates:

plot_phasor(
    real_filtered,
    imag_filtered,
    frequency=frequency,
    title='Median-filtered phasor coordinates (3x3 kernel, 3 repetitions)',
)

# %%
# Increasing the number of repetitions or the filter kernel size can further
# reduce noise, but may also remove relevant features:

mean_filtered, real_filtered, imag_filtered = phasor_threshold(
    *phasor_filter_median(mean, real, imag, repeat=6, size=5), mean_min=1
)

plot_phasor(
    real_filtered,
    imag_filtered,
    frequency=frequency,
    title='Median-filtered phasor coordinates (5x5 kernel, 6 repetitions)',
)

# %%
# The smoothing effect of median-filtering is demonstrated by plotting the
# real components of the filtered and unfiltered phasor coordinates as images:


def plot_images(im0, im1):
    # TODO: replace with future phasorpy.plot.plot_images
    from matplotlib import pyplot

    fig, (ax0, ax1) = pyplot.subplots(
        1, 2, figsize=(6.4, 3.4), subplot_kw={'xticks': [], 'yticks': []}
    )
    fig.suptitle('Real component of phasor coordinates')
    ax0.imshow(im0, vmin=0.4, vmax=0.9)
    ax0.set_title('Unfiltered')
    ax1.imshow(im1, vmin=0.4, vmax=0.9)
    ax1.set_title('Median-filtered')
    pyplot.tight_layout()


plot_images(phasor_threshold(mean, real, imag, mean_min=1)[1], real_filtered)

# %%
# pawFLIM wavelet filter
# ----------------------
#
# Filtering based on wavelet decomposition is another method to reduce noise.
# The function :py:func:`phasorpy.phasor.phasor_filter_pawflim` is based
# on the `pawFLIM <https://github.com/maurosilber/pawflim>`_ library.
# While the median filter is applicable to any type of phasor coordinates,
# the pawFLIM filter requires calibrated phasor coordinates from FLIM
# measurements and at least one harmonic and its corresponding double:

harmonic = [1, 2]

mean, real, imag = phasor_from_signal(signal, axis=0, harmonic=harmonic)
reference_mean, reference_real, reference_imag = phasor_from_signal(
    reference_signal, axis=0, harmonic=harmonic
)

real, imag = phasor_calibrate(
    real,
    imag,
    reference_mean,
    reference_real,
    reference_imag,
    frequency=frequency,
    lifetime=4.2,
    harmonic=harmonic,
)

# %%
# Apply the pawFLIM wavelet filter to the calibrated phasor coordinates:

mean_filtered, real_filtered, imag_filtered = phasor_threshold(
    *phasor_filter_pawflim(mean, real, imag, harmonic=harmonic), mean_min=1
)

# %%
# Plot the pawFLIM-filtered and thresholded phasor coordinates:

plot_phasor(
    real_filtered[0],
    imag_filtered[0],
    frequency=frequency,
    title='pawFLIM-filtered phasor coordinates for first harmonic',
)

# %%
# Increasing the significance level of the comparison between phasor
# coordinates and/or the maximum averaging area can further reduce noise:

mean_filtered, real_filtered, imag_filtered = phasor_filter_pawflim(
    mean, real, imag, harmonic=harmonic, sigma=5, levels=3
)
mean_filtered, real_filtered, imag_filtered = phasor_threshold(
    mean_filtered, real_filtered, imag_filtered, 1
)

plot_phasor(
    real_filtered[0],
    imag_filtered[0],
    frequency=frequency,
    title='pawFLIM-filtered phasor coordinates for first harmonic',
)

# %%
# sphinx_gallery_thumbnail_number = -1
# mypy: allow-untyped-defs, allow-untyped-calls

```

cgohlke:
`pawflim` needs to be added to https://github.com/phasorpy/phasorpy/blob/3c04dca9213df35d77dec06041b81a875b015bcf/pyproject.toml#L191

src/phasorpy/phasor.py
```diff
@@ -3026,6 +3028,209 @@ def phasor_filter_median(
...
+    real = numpy.asarray(real)
+    imag = numpy.asarray(imag)
+
+    if levels < 0:
+        raise ValueError(f'{levels=} < 0')
+    if levels == 0:
+        return mean, real, imag
+
+    if mean.shape != real.shape[-mean.ndim if mean.ndim else 1 :]:
+        raise ValueError(f'{mean.shape=} != {real.shape=}')
```
cgohlke, line 3145: Can you add a test for this? Should be trivial.

tutorials/api/phasorpy_filtering.py
```diff
@@ -0,0 +1,206 @@
...
+)
+mean_filtered, real_filtered, imag_filtered = phasor_threshold(
+    mean_filtered, real_filtered, imag_filtered, 1
+)
+
+plot_phasor(
+    real_filtered[0],
+    imag_filtered[0],
+    frequency=frequency,
+    title='pawFLIM-filtered phasor coordinates for first harmonic',
```
cgohlke, line 200: `title='pawFLIM-filtered phasor coordinates (sigma=5, levels=3)'`

tutorials/api/phasorpy_filtering.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+# %%
+# Plot the pawFLIM-filtered and thresholded phasor coordinates:
+
+plot_phasor(
+    real_filtered[0],
+    imag_filtered[0],
+    frequency=frequency,
+    title='pawFLIM-filtered phasor coordinates for first harmonic',
```
cgohlke, line 182: `title='pawFLIM-filtered phasor coordinates (sigma=2, levels=1)'`

cgohlke (approved):

cgohlke:
This looks good to me. My main concern is still the dependency on the `pawflim` package.

cgohlke:
Just noticed: maybe the tutorial section "Plot unfiltered phasor coordinates" could be named "No filter" or "Unfiltered" instead?

bruno-pannunzio:
> Just noticed: maybe the tutorial section "Plot unfiltered phasor coordinates" could be named "No filter" or "Unfiltered" instead?

Good idea. Previously it was kind of repetitive. Do you think is now ready for merging?

cgohlke:
> Do you think is now ready for merging?

Yes.

PR #172, opened 2025-01-12
Support reading point mode PTU files

cgohlke:
## Description

This PR adds support for reading point mode PTU files, commonly used to store instrument response functions, to the `read_ptu` function.

Items to discuss:

- No public sample file is available for testing, so coverage in CI decreases.
- The T, Y, and X dimensions are prepended to the point mode TCSPC histograms to make them broadcastable with image data returned by the function. That should facilitate calibration with IRF.
- The xarray returned by the `read_ptu` function now has an additional `ptu_tags` attribute, which contains the original tags read from PTU file. I'm not sure this is a good idea as it adds support overhead, and the metadata might be out of sync with the returned data. If yes, we might want to consider returning original metadata for other read functions as well. Should the attribute have a generic name like `original_metadata` or `vendor_metadata`? But then, users might expect it in a common format, like dict, not xml or numpy recarray...

## Checklist

### DIFF
--- a/src/phasorpy/io.py
+++ b/src/phasorpy/io.py
@@ -1229,7 +1229,7 @@ def read_ptu(
     dtime: int | None = 0,
     keepdims: bool = True,
 ) -> DataArray:
-    """Return image histogram and metadata from PicoQuant PTU T3 mode file.
+    """Return TCSPC histogram and metadata from PicoQuant PTU T3 mode file.
 
     PTU files contain time-correlated single photon counting measurement data
     and instrumentation parameters.
@@ -1239,7 +1239,7 @@ def read_ptu(
     filename : str or Path
         Name of PTU file to read.
     selection : sequence of index types, optional
-        Indices for all dimensions:
+        Indices for all dimensions of image mode files:
 
         - ``None``: return all items along axis (default).
         - ``Ellipsis``: return all items along multiple axes.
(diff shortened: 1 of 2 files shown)


bruno-pannunzio (approved):
Hi @cgohlke! This looks good to me although I can't test it because I also don't have IRF files. But this is very nice to have, I think many users (thinking of Leica also maybe?) will use calibration with IRF, which we currently don't support. We may discuss in a future PR/issue how to modify `phasor_calibrate` handling calibration with IRF.

 > * No public sample file is available for testing, so coverage in CI decreases.
 
I think we can ask Picoquant for this, they are very open with this.

> * The T, Y, and X dimensions are prepended to the point mode TCSPC histograms to make them broadcastable with image data returned by the function.

This is also a good idea, maybe we can revisit this when we modify `phasor_calibrate` if needed.

> * If yes, we might want to consider returning original metadata for other read functions as well. Should the attribute have a generic name like `original_metadata` or `vendor_metadata`? But then, users might expect it in a common format, like dict, not xml or numpy recarray...

I agree, I think this would be a great addition, although it may require some work. It would be nice to have all the available metadata in some way, to avoid losing all this information when going through different ways of analyzing data. For example, reading the file with PhasorPy, and exporting the data to open it in, lets say, ImageJ. It would be nice not to lose the metadata along the way. My guess is as you said that the metadata should be consistent for all file formats, maybe there is some way of parsing the information into a common format? But as I said, I guess this requires a lot of work and may not be possible for all file formats. We can discuss this further in out meetings if it makes sense.

cgohlke:
Thanks. Let's discuss metadata in the next meeting. Evangelos also mentioned the issue.

PR #189, opened 2025-02-14
Remove notes about early stage of development

cgohlke:
## Description

Remove notes about early stage of development.

## Checklist

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -4,6 +4,10 @@ API reference
 This documents the application programming and command line interfaces of the
 PhasorPy library version |version|.
 
+.. note::
+    The PhasorPy library is still under construction. Backwards-incompatible
+    changes may occur between revisions.
+
 .. toctree::
     :maxdepth: 3
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -4,11 +4,6 @@ API reference
 This documents the application programming and command line interfaces of the
 PhasorPy library version |version|.
 
-.. note::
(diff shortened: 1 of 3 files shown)


bruno-pannunzio:
Hi @cgohlke, I think it's a good idea to update since we are not in the early stages anymore. But I am wondering if it would be better to leave a note that still some large changes can occur, at least until we have the first release after version 1.x? I don't know if this is what we planned, but I think some big changes (such as the `phasor_filter` recently), can still occur. Do you think this is possible or we are at the point were we should not do this big changes?

cgohlke:
Hi @bruno-pannunzio : I agree. I was hoping that the current version numbering scheme (`0.x`) is sufficient to signal that these revisions are not necessarily backwards-compatible.

How about using:
```
.. note::
    The PhasorPy library is still under construction: backwards-incompatible
    changes may occur between revisions.
```

bruno-pannunzio:
> I was hoping that the current version numbering scheme (`0.x`) is sufficient to signal that these revisions are not necessarily backwards-compatible.

I think for most people yes, but it may be not too obvious for beginners and/or non programmers specially.

> .. note::
>     The PhasorPy library is still under construction: backwards-incompatible
>     changes may occur between revisions.

I like this very much! Very clear and on point.

cgohlke:
I added the tuned-down notes back to the API reference and Release notes, but not the documentation main page.

Let's apply large changes (like removing the `.phasor` namespace or renaming `components` functions) before the next 0.5 release and then try to more stable.

Issue #200, opened 2025-03-05
Cursor segmentation using machine learning


R-Blanco802:
It was observed that when a user of the PhasorPy library has obtained the phasors and wants to pseudo-color the image, they will need to manually analyze the distribution of the points in the phasor space in order to select which ROI will be coloured.
This may become cumbersome if they have a large number of images to analyze, and may also introduce biases in users. For example, a user who is trained on certain type of phasor images may tend to extract similar observations in the future, even when analyzing differently distributed phasor points which could lead to misinterpretation.

This issue proposes to automatize the observation of the distribution of phasor points, and selecting the cursors (further corrections are placed on the user, and how does the pseudo-coloured image looks like to them specifically).

In order to do this, I suggest using some scikit-learn (sklearn) modules such as the GMM (Gaussian Mixture Models) or k-means to find clusters in the phasor image and use them to choose centers, radii (in case of elliptical cursors), etc. for the cursors.

I am preparing a Pull Request for this, but I wanted your input on how the module's functions should operate and how to address certain problems: 
- One challenge is handling NaN values, as they pose a problem for sklearn. This was temporarily mitigated by replacing NaNs with zeros. 
- Another complication arises when two cursors overlap in the same region causing images to loose colour. For instance, if a blue and a red cursor are present, the expected result would be a mix of both with purple in the middle. However, the image appears predominantly gray with some blue, red and purple.

Thanks in advance!

bruno-pannunzio:
Hi @R-Blanco802! Thanks for showing interest in the PhasorPy project and for contributing to it. The automatic clustering is something we are very interested in having in the library, so this is much welcome.

> * One challenge is handling NaN values, as they pose a problem for sklearn. This was temporarily mitigated by replacing NaNs with zeros.

The `NaN` handling is something we are aiming for the library, since many functions return `NaN` for invalid values (such as an invalid fourier transformation or after thresholding). This can be a problem, particularly when using other libraries as 'sklearn'. The problem with replacing `NaN` with zeroes is that this are valid values for phasor coordinates and can affect the analysis if not handled properly. Maybe when you open the PR, we can discuss based on your implementation what can be an efficient way to handle the `NaN` without skewing the analysis. This can be done through masks or maybe other alternatives.

> * Another complication arises when two cursors overlap in the same region causing images to loose colour. For instance, if a blue and a red cursor are present, the expected result would be a mix of both with purple in the middle. However, the image appears predominantly gray with some blue, red and purple.

[...]

Issue #158, opened 2024-12-09
Static code analysis errors with numpy 2.2.0


cgohlke:
The "Static code analysis" action currently fails with several mypy errors, most probably due to the release of [numpy 2.2](https://numpy.org/devdocs/release/2.2.0-notes.html) yesterday, which contains "Many improved annotations."

For example https://github.com/phasorpy/phasorpy/actions/runs/12243727085/job/34153909334:

```
Run python -m mypy
src/phasorpy/color.py:63: error: Incompatible return value type (got "tuple[Any | float | list[float] | list[list[Any]], Any | float | list[float] | list[list[Any]], Any | float | list[float] | list[list[Any]]]", expected "tuple[float, float, float] | ndarray[tuple[int, ...], dtype[Any]]")  [return-value]
src/phasorpy/color.py:63: error: Value of type "float | list[float] | list[list[float]] | list[list[list[Any]]]" is not indexable  [index]
src/phasorpy/utils.py:184: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[Any]]", variable has type "ndarray[tuple[int, int], dtype[Any]]")  [assignment]
src/phasorpy/utils.py:186: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[Any]]", variable has type "ndarray[tuple[int, int], dtype[Any]]")  [assignment]
src/phasorpy/phasor.py:703: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[float64]]", variable has type "ndarray[tuple[int], dtype[float64]]")  [assignment]
src/phasorpy/phasor.py:2133: error: Incompatible types in assignment (expression has type "ndarray[tuple[int, ...], dtype[float64]]", variable has type "ndarray[tuple[int, int, int], dtype[float64]]")  [assignment]
src/phasorpy/plot.py:340: error: No overload variant of "__setitem__" of "ndarray" matches argument types "Any", "None"  [call-overload]
src/phasorpy/plot.py:340: note: Possible overload variants:
src/phasorpy/plot.py:340: note:     def __setitem__(self, str | list[str], Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | bool | int | float | complex | str | bytes | _NestedSequence[bool | int | float | complex | str | bytes], /) -> None
[...]

PR #201, opened 2025-03-05
Add phasor_cluster_gmm function

R-Blanco802:
## Description

Clustering module with function to perform gaussian mixture models, returning the centers, radii, etc. of the selected amount of clusters, in order to tackle issue #200 .
This functions will work on the returns of other functions from the repository, although not with good results on the pseudo-coloured images.

I have added (and tried to clean the code I use to try out the function) on phasorpy/tutorial/api and it is called phasorpy_clustering.py 
The module is located on phasorpy/src, and is called clustering.py

## Checklist


cgohlke:
Hi @R-Blanco802 To facilitate code review, can you please run `pre-commit` on your code and make sure the documentation and tutorial build without errors. See [here](https://www.phasorpy.org/docs/stable/contributing/#code-standards) and [here](https://www.phasorpy.org/docs/stable/contributing/#documentation).

R-Blanco802:
Hi @cgohlke, sorry about that. 
This fix should now pass the pre-commit test. However I am having some trouble with mypy over sklearn, which form what I have read is an issue over their repo [https://github.com/scikit-learn/scikit-learn/issues/16705#issuecomment-2421887710]. And mypy is also showing some errors at plot.py and phasor.py which I haven't touched, maybe I am forgetting to add some library?

Documentations test fail on 7 files, these are: _io.py, components.py, cursors.py, io.py, phasor.py, plot.py and utils.py. I don't really understand documentation tests with pytest, but I'll try to learn it.

cgohlke:
Thanks @R-Blanco802 : the Mypy errors you get are probably due to issue #158. Numpy 2.2.x has someproblems with mypy, so we validate typing only with numpy 2.1.

Here are some quick fixes for the remaining errors I get on my system. After those are applied, workflows can be enabled for this PR.

```diff
diff --git a/pyproject.toml b/pyproject.toml
index a6c899c..098c615 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,7 +144,7 @@ strict = true
 warn_unreachable = true

 [[tool.mypy.overrides]]
-module = ["phasorpy._phasorpy", "pooch.*", "scipy.*", "mkl_fft.*"]
+module = ["phasorpy._phasorpy", "pooch.*", "scipy.*", "mkl_fft.*", "sklearn.*"]
 ignore_missing_imports = true

 [tool.coverage.run]
diff --git a/src/phasorpy/clustering.py b/src/phasorpy/clustering.py
index bce4bf8..d06ff83 100644
--- a/src/phasorpy/clustering.py
+++ b/src/phasorpy/clustering.py
@@ -67,7 +67,7 @@ def gaussian_mixture_model(
     >>> component_lifetimes = [1.0, 8.0]
     >>> component_fractions = [0.7, 0.3]
     >>> frequency = 80.0  # MHz
-
+    >>> from phasorpy.phasor import phasor_from_lifetime
     >>> real, imag = phasor_from_lifetime(
     ...     frequency, component_lifetimes, component_fractions
     ... )
diff --git a/tutorials/api/phasorpy_clustering.py b/tutorials/api/phasorpy_clustering.py
index 0953b0b..7bfaf09 100644
--- a/tutorials/api/phasorpy_clustering.py
+++ b/tutorials/api/phasorpy_clustering.py
@@ -1,6 +1,6 @@
 """
 Clustering
-=======
+==========

 An introduction to clustering phasor points to be used with cursors.

@@ -12,9 +12,7 @@ An introduction to clustering phasor points to be used with cursors.
 import matplotlib.pyplot as plt
 import numpy

-# import clustering # To be solved!!!
-from clustering import gaussian_mixture_model
-
+from phasorpy.clustering import gaussian_mixture_model
 from phasorpy.color import CATEGORICAL
 from phasorpy.cursors import (
     mask_from_circular_cursor,
@@ -72,9 +70,6 @@ mean_filtered, real_filtered, imag_filtered = phasor_threshold(
 # ====================clustering======================
 # ====================================================

-import clustering
-from clustering import gaussian_mixture_model
-
 n_components = 2

 centers_real, centers_imag, radio, radius_minor, angles = (
```

cgohlke:
I'll leave the in-depth code review to @bruno-pannunzio but have some comments:

- I would prefer the module and tutorial to be named `cluster` (instead of `clustering`) and the function `phasor_cluster_gmm` (similar to `phasor_filter_median`)
- `scikit-learn` needs to be added as a requirement at https://github.com/phasorpy/phasorpy/blob/8fd32048b52586c0e86db3655c40e27c5c320fd4/pyproject.toml#L24, https://github.com/phasorpy/phasorpy/blob/8fd32048b52586c0e86db3655c40e27c5c320fd4/pyproject.toml#L192, https://github.com/phasorpy/phasorpy/blob/8fd32048b52586c0e86db3655c40e27c5c320fd4/requirements_dev.txt#L58, https://github.com/phasorpy/phasorpy/blob/8fd32048b52586c0e86db3655c40e27c5c320fd4/requirements_min.txt#L8
- The module needs to be added to the documentation at https://github.com/phasorpy/phasorpy/tree/main/docs/api and https://github.com/phasorpy/phasorpy/blob/main/docs/api/index.rst
- Replacing NaNs by 0 won't work since (0,0) are valid phasor coordinates. Does replacing NaNs with 1 or other coordinates way outside the unit circle work?
- Tests are missing (OK for now)

bruno-pannunzio:
Hi @R-Blanco802, this looks very nice. I will be adding some comments and suggestions first to improve the `phasor_cluster_gmm` function and then we can work on the tutorial and tests.

I think some of the parts of the function can be simplified and optimized. Particularly the `NaN` handling and all the reshaping can be reduced. I believe we can use masks to ignore the `NaN` in the fitting of the model and avoid creating a separate cluster with coordinates (2,2). I will add this comment in line with the code for easier comparison. I haven't tested it thoroughly but I think this can work. 

Also, I don't think the function supports arrays with axes that need to be skipped for clustering. For example, FLIM data with with harmonic axis, or different axes for frequencies or channels, that need to be clustered separately for each harmonic, channel, etc. Some of the function in the library have a `skip_axis` parameter to handle this situations and specifying which axes to ignore. I think this function needs this parameter, if not all axes will be flattened and the clustering will not be correct. Another alternative would be to explicitly specify that this function doesn't work in this conditions, but I believe with minor adjustments that can be done.

Finally, I think the angle calculations are not correct. From the tutorial you can see that the ellipses are shifted. I have tested it with other data and it is also the case. I will suggest a comment for fixing this in the code that I have tried in a few examples and it works.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+from typing import List, Tuple  # Needed to correct some mypy issues
+
+import numpy
+from sklearn.mixture import GaussianMixture
+
+
+def phasor_cluster_gmm(
+    real: ArrayLike,
+    imag: ArrayLike,
+    n_components: int,
```
bruno-pannunzio, line 22: I would suggest to change the paramater `n_components` to `clusters` or `n_clusters`. I personally prefer `clusters`, but we can discuss this. I think components can be misleading in the case of phasor analysis since we use it for specifying the number of fluorescent species, which in this case it may not be the same as the cluster of points. The clusters depend on the fraction of this components in the pixels.

bruno-pannunzio: I see that the components comes from the mixture models, but still, I think clusters is better.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+    **kwargs: Any,
```
bruno-pannunzio, line 23: The `**kwargs` parameter should also be added to the description in the docstring of the function. You can look for other function in the library that use `**kwargs` as parameter to pass parameter to other functions and use the same logic in the docstring.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+        The numbers of components present in a sample, i.e. fluorophores
+        that will be present, or are thought to be present.
+
+    Returns
+    -------
+    centers_real : array_like, shape (n_components,)
+        Real coordinates of ellipses centers.
+    centers_imag : array_like, shape (n_components,)
+        Imaginary coordinates of ellipses centers.
+    radio : array_like, shape (n_components,)
```
bruno-pannunzio, line 44: 'Radius' instead of 'radio'

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+    # TODO: See Also (I would need the link).
+    Raises
+    ------
+    ValueError
+        The array shapes of `real` and `imag` do not match.
+        n_components is not an integer.
+
+    See Also
+    --------
+    :ref:`sphx_glr_tutorials_api_phasorpy_cursors.py`
```
bruno-pannunzio, line 61: I don't think the cursors tutorial should be referenced here. Maybe in the tutorial.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+from sklearn.mixture import GaussianMixture
+
+
+def phasor_cluster_gmm(
+    real: ArrayLike,
+    imag: ArrayLike,
+    n_components: int,
+    **kwargs: Any,
+) -> tuple[list[Any], list[Any], list[float], list[float], list[Any]]:
+    """Returns centers, radii and angles of clusters from
```
bruno-pannunzio, line 25: The description of the function should start with a single line. Maybe something like:

`Return clusters from phasor coordinates using Gaussian Mixture Model.`

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+        Imaginary coordinates of ellipses centers.
+    radio : array_like, shape (n_components,)
+        Radii of ellipses along semi-major axis.
+    radius_minor : array_like, shape (n_components,)
+        Radii of ellipses along semi-minor axis.
+    angle : array_like, optional, shape (n_components,)
+        Rotation angles of semi-major axes of ellipses in radians.
+
+
+    # TODO: See Also (I would need the link).
```
bruno-pannunzio, line 52: I think here you are refering to the link of the paper:
https://doi.org/10.1364/BOE.422766

You can reference it like in this other function from the library:
https://www.phasorpy.org/docs/dev/api/phasor/#phasorpy.phasor.phasor_filter_pawflim

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,209 @@
...
+        u = eigenvectors[0] / numpy.linalg.norm(eigenvectors[0])
+
+        angle = numpy.arctan2(u[1], u[0])
+        angle = 180 * angle / numpy.pi  # convert to degrees
+        eigenvalues = 2.0 * numpy.sqrt(2.0) * numpy.sqrt(eigenvalues)
+        angle = math.radians(angle)  # convert to radians
+
+        angles = numpy.vstack([angles, [angle]])
+        radius = numpy.vstack([radius, eigenvalues])
+
```
bruno-pannunzio, line 161: I tried this changes in the loop for fixing the angles of the ellipsis. Maybe you can try it and adapt it for the rest of the changes in the funciton:

```Python

    for n in range(n_components):
        cov = gm.covariances_[n][:2, :2]
        eigenvalues, eigenvectors = numpy.linalg.eigh(cov)
        
        # Get the index of the largest eigenvalue (major axis)
        idx = numpy.argmax(eigenvalues)
        
        # Use the eigenvector corresponding to the largest eigenvalue
        major_axis = eigenvectors[:, idx]
        
        # Calculate angle using arctan2
        angle = numpy.arctan2(major_axis[1], major_axis[0])
        
        # Ensure the angle is in the range [0, π]
        if angle < 0:
            angle += numpy.pi
            
        # Scale eigenvalues for confidence ellipse
        eigenvalues = 2.0 * numpy.sqrt(2.0) * numpy.sqrt(eigenvalues)
        
        angles = numpy.vstack([angles, [angle]])
        radius = numpy.vstack([radius, eigenvalues])

```

src/phasorpy/cluster.py
R-Blanco802: Sorry about that, I was trying to figure out what  this meant     :ref:`sphx_glr_tutorials_api_phasorpy_cursors.py`, and found that it was a reference to phasorpy's tutorial about cursors.

I believe that changing line 61 to :ref:`sphx_glr_tutorials_api_phasorpy_cluster.py`
Should make the correct reference to the cluster tutorial. Is that right?

R-Blanco802:
Thank you for the feedback. The proposed solution is much cleaner.
The angle-based solution is now working as expected. 

However, I have been considering the implementation of skip_axis, which I believe will improve the module's cohesion within the project. 
This has not been implemented yet, but I am currently working on it. Aside from that, all other issues should now be resolved, unless something was overlooked.

bruno-pannunzio:
Hi @R-Blanco802, the modifications seem to be working great. I think now is much cleaner the implementation. I have a few more suggestions that I will be making in line for easier comparison.

I will also suggest some changes to the tutorial to have the same format as the other tutorials.

The implementation of the `skip_axis` parameter make be a little tricky, so take your time and let us know if we can be of help.

I enabled the workflows so that we can start checking if tests are passing. Everything seems to be working fine. Since the tests are missing the codecoverage test will give an error since the new added lines are not tested. Once the tests are implemented, this should be OK. There is one test that is failing but is related to some modification in tests files that we are implementing in another PR, so don't consider the test that is failing for now.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,181 @@
+"""Clustering techniques."""
```
bruno-pannunzio, line 1: I would suggest to expand a little bit more on the description of the module and the functions included. For example:

"""Cluster phasor coordinates.

The ``phasorpy.cluster`` module provides functions to:

- cluster phasor coordinates using Gaussian Mixture Model (GMM).

  - :py:func:`phasor_cluster_gmm`

"""

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,181 @@
...
+from __future__ import annotations
+
+__all__ = ['phasor_cluster_gmm']
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ._typing import Any, ArrayLike, NDArray
+
+from typing import List, Tuple  # Needed to correct some mypy issues
```
bruno-pannunzio, line 12: If these are not going to be used maybe they can be removed (if the MyPy errors were solved)

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,181 @@
...
+    imag = numpy.asarray(imag)
+
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+    if not isinstance(clusters, int):
+        raise ValueError(
+            f"clusters is expected to be an integer, but got {clusters=} of type {type(clusters)}"
+        )
+
+    # Reshape and stack the coordinate
```
bruno-pannunzio, line 112: I would remove all these comments unless they help to understand something not directly visible in the code.

src/phasorpy/cluster.py
cgohlke: What other functions could be added to the `cluster` module (to justify creating a new module)?

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,181 @@
...
+        angles = numpy.vstack([angles, [angle]])
+        radio = numpy.vstack([radio, eigenvalues])
+
+    # Generate the empty arrays
+    centers_real = numpy.array([]).reshape(0, 1)
+    centers_imag = numpy.array([]).reshape(0, 1)
+
+    for n in range(clusters):
+        centers_real = numpy.vstack([centers_real, [gm.means_[n, 0]]])
+        centers_imag = numpy.vstack([centers_imag, [gm.means_[n, 1]]])
```
bruno-pannunzio, line 155: I don't know how the `skip_axis` parameter will be implemented, but I think it makes more sense for the function to return a tuple of numpy arrays instead of list. I am not sure, we can discuss this, but I think it makes sense considering how it will handle when axes are skipped, it should return an array with a shape corresponding to the axes excluded.

src/phasorpy/cluster.py
bruno-pannunzio: That is a good question Christoph. There are other clustering techniques that can be used, such as 'k-means' or 'k-mediods', etc. I don't know if we plan/want to add this to the library. If this is the only function for clustering (also considering while no other clustering functions are implemented), in which other module can this function be included? The `phasor` module maybe? It is getting a little bit crowded but one more function won't hurt.

R-Blanco802: I was thinking about adding in the future an implementation of k-means, also using scikit learn.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+
+    real = numpy.asarray(real)
+    imag = numpy.asarray(imag)
+
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+    if not isinstance(clusters, int):
+        raise ValueError(
+            f"clusters is expected to be an integer, but got {clusters=} of type {type(clusters)}"
```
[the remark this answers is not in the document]
cgohlke: Better use `isinstance(clusters, numbers.Integral)` to check for integers. That includes numpy scalars, for example, `numpy.int32(1)`

cgohlke:
The code can be streamlined, especially the repeated appending of values via `numpy.vstack`.

How about something like (with help from Claude Sonnet):

```Python
def phasor_cluster_gmm(
    real: ArrayLike,
    imag: ArrayLike,
    /,
    *,
    clusters: int = 1,
    **kwargs: Any,
) -> tuple[
    tuple[float, ...],
    tuple[float, ...],
    tuple[float, ...],
    tuple[float, ...],
    tuple[float, ...],
]:
    """Return elliptic clusters in phasor coordinates using GMM.

    Fit a Gaussian Mixture Model (GMM) to the provided phasor coordinates and
    extract the parameters of ellipses that represent each cluster according
    to [1]_.

    Parameters
    ----------
    real : array_like
        Real component of phasor coordinates.
    imag : array_like
        Imaginary component of phasor coordinates.
    clusters : int, optional
        Number of Gaussian distributions to fit to phasor coordinates.
        Defaults to 1.
    **kwargs
        Additional keyword arguments passed to
        `sklearn.mixture.GaussianMixture`.
        Common options include:

        - covariance_type : {'full', 'tied', 'diag', 'spherical'}
        - max_iter : int, maximum number of EM iterations
        - random_state : int, for reproducible results

    Returns
    -------
    center_real : tuple of float
        Real component of ellipse centers.
    center_imag : tuple of float
        Imaginary component of ellipse centers.
    radius_major : tuple of float
        Major radii of ellipses.
    radius_minor : tuple of float
        Minor radii of ellipses.
    angle : tuple of float
        Rotation angles of major axes in radians, within range [0, π].

    Raises
    ------
    ValueError
        If the array shapes of `real` and `imag` do not match.
        If `clusters` is not a positive integer.

    Notes
    -----
    The radii represent the 95% confidence intervals of the Gaussian
    distributions, scaled by 2.0 * sqrt(2.0) * sqrt(eigenvalues).

    See Also
    --------
    :ref:`sphx_glr_tutorials_api_phasorpy_cluster.py`

    References
    ----------
    .. [1] Vallmitjana A, Torrado B, and Gratton E.
      `Phasor-based image segmentation: machine learning clustering techniques
      <https://doi.org/10.1364/BOE.422766>`_.
      *Biomed Opt Express*, 12(6): 3410-3422 (2021).

    Examples
    --------
    Recover the clusters from a synthetic distribution of phasor coordinates
    with two clusters:

    >>> real1, imag1 = numpy.random.multivariate_normal(
    ...     [2, 3], [[0.3, 0.1], [0.1, 0.2]], 100
    ... ).T
    >>> real2, imag2 = numpy.random.multivariate_normal(
    ...     [5, 6], [[0.2, -0.1], [-0.1, 0.3]], 100
    ... ).T
    >>> real = numpy.concatenate([real1, real2])
    >>> imag = numpy.concatenate([imag1, imag2])
    >>> (center_real, center_imag, radius_major, radius_minor, angle) = (
    ...     phasor_cluster_gmm(real, imag, clusters=2)
    ... )

    """
    real_array = numpy.asarray(real)
    imag_array = numpy.asarray(imag)

    # validate inputs
    if real_array.shape != imag_array.shape:
        raise ValueError(f"{real_array.shape=} != {imag_array.shape=}")

    if (
        not isinstance(
            clusters, (int, numbers.Integral)
        )  # type: ignore[redundant-expr]
        or clusters < 1
    ):
        raise ValueError(
            f'clusters must be a positive integer, got {clusters}'
        )
    clusters = int(clusters)

    # prepare data for clustering
    coords = numpy.column_stack((real_array.ravel(), imag_array.ravel()))

    # remove NaN values if present
    valid_data = ~numpy.isnan(coords).any(axis=1)
    coords = coords[valid_data]

    if coords.shape[0] < clusters:
        raise ValueError(
            f'not enough valid data points ({coords.shape[0]}) for '
            f'requested number of components ({clusters})'
        )

    kwargs.pop('n_components', None)

    # fit Gaussian Mixture Model
    gmm = GaussianMixture(n_components=clusters, n_init=10, **kwargs)
    gmm.fit(coords)

    # extract ellipse parameters
    center_real = []
    center_imag = []
    radius_major = []
    radius_minor = []
    angle = []

    for i in range(clusters):
        # center coordinates
        center_real.append(float(gmm.means_[i, 0]))
        center_imag.append(float(gmm.means_[i, 1]))

        # get covariance matrix
        if gmm.covariance_type == 'full':
            cov = gmm.covariances_[i]
        elif gmm.covariance_type == 'tied':
            cov = gmm.covariances_
        elif gmm.covariance_type == 'diag':
            cov = numpy.diag(gmm.covariances_[i])
        else:  # 'spherical'
            cov = numpy.eye(2) * gmm.covariances_[i]

        # Eigen decomposition
        eigenvalues, eigenvectors = numpy.linalg.eigh(cov[:2, :2])

        # sort eigenvalues and eigenvectors in descending order
        idx = eigenvalues.argsort()[::-1]
        eigenvalues = eigenvalues[idx]
        eigenvectors = eigenvectors[:, idx]

        # calculate angle from major eigenvector
        major_vector = eigenvectors[:, 0]
        current_angle = numpy.arctan2(major_vector[1], major_vector[0])

        # ensure angle is in [0, π] range
        if current_angle < 0:
            current_angle += numpy.pi

        angle.append(float(current_angle))

        # scale eigenvalues for 95% confidence ellipse (2σ)
        scaling = 2.0 * math.sqrt(2.0)
        radius_major.append(scaling * math.sqrt(eigenvalues[0]))
        radius_minor.append(scaling * math.sqrt(eigenvalues[1]))

    return (
        tuple(center_real),
        tuple(center_imag),
        tuple(radius_major),
        tuple(radius_minor),
        tuple(angle),
    )
```

R-Blanco802:
I will try to implement the streamlined version, some comments on the code:

>         # scale eigenvalues for 95% confidence ellipse (2σ)
>         scaling = 2.0 * math.sqrt(2.0)
>         radius_major.append(scaling * math.sqrt(eigenvalues[0]))
>         radius_minor.append(scaling * math.sqrt(eigenvalues[1]))

I had been experimenting with a scaling factor to address the problem of cursor superposition. However, I didn’t include it in the function since I was mainly forcing it to work as a temporary fix. If you think it would be a useful addition, I will incorporate it.

>         # get covariance matrix
>         if gmm.covariance_type == 'full':
>             cov = gmm.covariances_[i]
>         elif gmm.covariance_type == 'tied':
>             cov = gmm.covariances_
>         elif gmm.covariance_type == 'diag':
>             cov = numpy.diag(gmm.covariances_[i])
>         else:  # 'spherical'
>             cov = numpy.eye(2) * gmm.covariances_[i]
> 
>         # Eigen decomposition
>         eigenvalues, eigenvectors = numpy.linalg.eigh(cov[:2, :2])

Regarding covariance_type, I didn’t fully understand its role, which is why I didn’t include any manipulation of it in the function. However, based on the code you provided, it appears to be important as it affects the shape of the covariance_ parameter and how it should be handled.

Lastly, n_init is indeed included in kwargs and defaults to 1. You can also modify how the weights are initialized by adjusting init_params. For example, using k-means++ provides a more efficient way to initialize k-means, leading to better clustering performance.

cgohlke:
The `covariance_type` handling came from the LLM. Obviously it needs to be tested.

`n_init` is currently fixed to 10.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,263 @@
...
+from collections.abc import Sequence
+
+import numpy
+from sklearn.mixture import GaussianMixture
+
+
+def phasor_cluster_gmm(
+    real: ArrayLike,
+    imag: ArrayLike,
+    scaling: float = 2 * numpy.sqrt(2),
```
cgohlke, line 29: Maybe its better to use `sigma=2.0` instead of `scaling` as a parameter. Sigma is already used in other functions.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,263 @@
...
+        tuple(center_imag),
+        tuple(radius_major),
+        tuple(radius_minor),
+        tuple(angle),
+    )
+
+
+# Added here, or should it be imported from phasor.py where
+# there's an exact copy of the function?
+def _parse_skip_axis(
```
cgohlke, line 207: `_parse_skip_axis` should be moved from `phasor.py` to `_utils.py` and imported from there. I'll open a PR.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,263 @@
...
+        major_vector = eigenvectors[:, 0]
+        current_angle = numpy.arctan2(major_vector[1], major_vector[0])
+
+        if current_angle < 0:
+            current_angle += numpy.pi
+
+        angle.append(float(current_angle))
+
+        radius_major.append(scaling * numpy.sqrt(eigenvalues[0]))
+        radius_minor.append(scaling * numpy.sqrt(eigenvalues[1]))
```
cgohlke, line 194: These are all scalar values, right? Better use functions and constants from the `math` module instead of numpy.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,263 @@
...
+    center_real : tuple of float
+        Real component of ellipse centers.
+    center_imag : tuple of float
+        Imaginary component of ellipse centers.
+    radius_major : tuple of float
+        Major radii of ellipses.
+    radius_minor : tuple of float
+        Minor radii of ellipses.
+    angle : tuple of float
+        Rotation angles of major axes in radians, within range [0, π].
```
cgohlke, line 87: So far, PhasorPy does not use Unicode characters in docstrings or code. Use `pi` instead of `π` or latex for math formulas.

src/phasorpy/cluster.py
cgohlke: If you update from main branch, you can import the function with `from ._utils import parse_skip_axis`

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+
+from ._utils import parse_skip_axis
+
+
+def phasor_cluster_gmm(
+    real: NDArray[numpy.float64],
+    imag: NDArray[numpy.float64],
+    sigma: float = 2.0,
+    /,
+    *,
```
cgohlke, line 33: Input arrays should be `array_like` and optional parameters specified by name.

```suggestion
def phasor_cluster_gmm(
    real: ArrayLike,
    imag: ArrayLike,
    /,
    *,
    sigma: float = 2.0,
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+    Parameters
+    ----------
+    real : array_like
+        Real component of phasor coordinates.
+    imag : array_like
+        Imaginary component of phasor coordinates.
+    sigma: float, default = 2.0
+        Scaling factor for the radii of major and minor axes. By default, it is set
+        to 2, which corresponds to the scaling of eigenvalues for a 95 per cent confidence
+        ellipse.
```
cgohlke, line 59:
```suggestion
    sigma: float, default = 2.0
        Scaling factor for radii of major and minor axes.
        Defaults to 2, which corresponds to the scaling of eigenvalues for a
        95% confidence ellipse.
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+    clusters : int, optional
+        Number of Gaussian distributions to fit to phasor coordinates.
+        Defaults to 1.
+    skip_axis : int or sequence of int, optional
+        Axes to skip in the data (useful for multi-dimensional arrays).
```
cgohlke, line 64:
```suggestion
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+    **kwargs
+        Additional keyword arguments passed to
+        :py:class:`sklearn.mixture.GaussianMixture`.
+
+        Common options include:
+        - covariance_type : {'full', 'tied', 'diag', 'spherical'}
+        - max_iter : int, maximum number of EM iterations
+        - random_state : int, for reproducible results
```
cgohlke, line 72: Add an empty line before the list.

```suggestion
        Common options include:

        - covariance_type : {'full', 'tied', 'diag', 'spherical'}
        - max_iter : int, maximum number of EM iterations
        - random_state : int, for reproducible results
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+    Raises
+    ------
+    ValueError
+        If the array shapes of `real` and `imag` do not match.
+        If `clusters` is not a positive integer.
+
+    Notes
+    -----
+    The radii represent the 95 per cent confidence intervals of the Gaussian
+    distributions, scaled by 2.0 * sqrt(2.0) * sqrt(eigenvalues).
```
cgohlke, line 96: I think this is outdated and can be removed.

```suggestion
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+
+    skip_axis, _ = parse_skip_axis(skip_axis, real.ndim)
+
+    if skip_axis:
+        real = numpy.mean(real, axis=skip_axis, keepdims=True)
+        imag = numpy.mean(imag, axis=skip_axis, keepdims=True)
+
+    coords: NDArray[numpy.float64] = numpy.column_stack(
+        (real.ravel(), imag.ravel())
+    )
```
cgohlke, line 141: This can be simplified a lot:

```suggestion
    coords = numpy.stack((real, imag), axis=-1).reshape(-1, 2)
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+
+__all__ = ['phasor_cluster_gmm']
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ._typing import Any, ArrayLike, NDArray
+
+import math
+from collections.abc import Sequence
```
cgohlke, line 20:
```suggestion
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+
+import numpy
+from sklearn.mixture import GaussianMixture
+
+from ._utils import parse_skip_axis
```
cgohlke, line 25:
```suggestion
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+"""
+
+from __future__ import annotations
+
+__all__ = ['phasor_cluster_gmm']
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ._typing import Any, ArrayLike, NDArray
```
cgohlke, line 17:
```suggestion
    from ._typing import Any, ArrayLike
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def phasor_cluster_gmm(
+    real: NDArray[numpy.float64],
+    imag: NDArray[numpy.float64],
+    sigma: float = 2.0,
+    /,
+    *,
+    clusters: int = 1,
+    skip_axis: int | Sequence[int] | None = None,
```
cgohlke, line 35: I don't understand the purpose of `skip_axis` in the current implementation and suggest to remove it.

```suggestion
```

tests/test_cluster.py
```diff
@@ -0,0 +1,119 @@
...
+    [
+        (numpy.array([[1, 2], [3, 4]]), numpy.array([[1, 2], [3, 4]])),
+        (numpy.array([1, 2, 3, 4]), numpy.array([1, 2, 3, 4])),
+    ],
+)
+def test_phasor_cluster_gmm_column_stack(real, imag):
+    centers_real, centers_imag, *_ = phasor_cluster_gmm(real, imag, clusters=1)
+
+    assert len(centers_real) == 1
+    assert len(centers_imag) == 1
```
cgohlke, line 119: Tests can be simplified:

```suggestion
"""Tests for the phasorpy.cluster module."""

import numpy
import pytest
from numpy.testing import assert_allclose

from phasorpy.cluster import phasor_cluster_gmm

numpy.random.seed(42)


@pytest.mark.parametrize('clusters', [1, 2, 3])
def test_phasor_cluster_gmm_basic(clusters):
    numpy.random.seed(42)
    real1, imag1 = numpy.random.multivariate_normal(
        [0.2, 0.3], [[0.003, 0.001], [0.001, 0.002]], 10000
    ).T
    real2, imag2 = numpy.random.multivariate_normal(
        [0.4, 0.5], [[0.002, -0.001], [-0.001, 0.003]], 10000
    ).T
    real = numpy.concatenate([real1, real2])
    imag = numpy.concatenate([imag1, imag2])

    centers_real, centers_imag, radius_major, radius_minor, angles = (
        phasor_cluster_gmm(real, imag, clusters=clusters)
    )

    assert len(centers_real) == clusters
    assert len(centers_imag) == clusters
    assert len(radius_major) == clusters
    assert len(radius_minor) == clusters
    assert len(angles) == clusters

    if clusters == 2:
        assert_allclose(centers_real, [0.4, 0.2], atol=0.01)
        assert_allclose(centers_imag, [0.5, 0.3], atol=0.01)
        assert_allclose(radius_major, [0.17, 0.17], atol=0.01)
        assert_allclose(radius_minor, [0.104, 0.104], atol=0.01)
        assert_allclose(angles, [2.114, 0.553], atol=0.01)


def test_phasor_cluster_gmm_invalid_shapes():
    with pytest.raises(ValueError):
        phasor_cluster_gmm([1, 2, 3], [1, 2])


@pytest.mark.parametrize(
    'covariance_type', ['full', 'tied', 'diag', 'spherical']
)
def test_phasor_cluster_gmm_covariance(covariance_type):
    centers_real, centers_imag, radius_major, radius_minor, angles = (
        phasor_cluster_gmm(
            [0.1, 0.2, 0.3],
            [0.1, 0.2, 0.3],
            clusters=2,
            covariance_type=covariance_type,
        )
    )

    assert len(centers_real) == 2
    assert len(centers_imag) == 2

    if covariance_type == 'full':
        assert isinstance(radius_major, tuple) and len(radius_major) == 2
    elif covariance_type == 'tied':
        assert isinstance(radius_major, tuple) and len(radius_major) == 2
    elif covariance_type == 'diag':
        assert isinstance(radius_major, tuple) and len(radius_major) == 2
    elif covariance_type == 'spherical':
        assert isinstance(radius_major, tuple) and len(radius_major) == 2


@pytest.mark.parametrize(
    'real, imag',
    [
        ([1, 2, 3], [1, 2]),
        ([1.0], [1.0]),
        ([1.0, numpy.nan, 2.0], [1.0, 2.0, numpy.nan]),
    ],
)
def test_phasor_cluster_gmm_exceptions(real, imag):
    with pytest.raises(ValueError):
        phasor_cluster_gmm(real, imag, clusters=2)


@pytest.mark.parametrize(
    'real, imag',
    [
        ([[1, 2], [3, 4]], [[1, 2], [3, 4]]),
        ([1, 2, 3, 4], [1, 2, 3, 4]),
    ],
)
def test_phasor_cluster_gmm_column_stack(real, imag):
    centers_real, centers_imag, *_ = phasor_cluster_gmm(real, imag, clusters=1)

    assert len(centers_real) == 1
    assert len(centers_imag) == 1


# mypy: allow-untyped-defs, allow-untyped-calls
# mypy: disable-error-code="arg-type, unreachable, redundant-expr"
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,193 @@
+"""Cluster phasor coordinates.
+
+The `phasorpy.cluster` module provides functions to:
+
+- cluster phasor coordinates using Gaussian Mixture Model (GMM):
+    - :py:func:`phasor_cluster_gmm`
```
cgohlke, line 6:
```suggestion
- fit elliptic clusters to phasor coordinates using
  Gaussian Mixture Model (GMM):

  - :py:func:`phasor_cluster_gmm`
```

cgohlke:
Hi @R-Blanco802 and @bruno-pannunzio : I left many suggestions in the review to finish this PR.

I think `skip_axes` should be left out of this implementation unless there is a convincing application that justifies complicating the code and output. Maybe a note could be added to the documentation that the function is usually intended to operate on single channel, single-harmonics phasor coordinates? Or am I missing something?

The tutorial needs to be polished to match the PhasorPy style. It would be nice to reproduce the results of the cited publication if the data files are public and not too large. For now, instead of adding a new tutorial, how about adding a short section to the introductory tutorial at https://www.phasorpy.org/docs/stable/tutorials/phasorpy_introduction/#spectral-phasors

src/phasorpy/cluster.py
cgohlke: Yes, mypy is a nuisance at times and needs to be worked around or ignored. But `NDArray[numpy.float64]` is really the wrong input type for a public function.

cgohlke:
I think the additions to the introduction tutorial are too detailed. All that is needed is a section "Find clusters" that seamlessly integrates with the existing spectral phasor section, for example:

```Python
# %%
# Spectral phasors
# ----------------
#
# Phasor coordinates can be calculated from hyperspectral images (acquired
# at many equidistant emission wavelengths) and processed in much the same
# way as time-resolved signals. Calibration is not necessary.
#
# Open a hyperspectral dataset acquired with a laser scanning microscope
# at 30 emission wavelengths:

from phasorpy.io import signal_from_lsm

hyperspectral_signal = signal_from_lsm(fetch('paramecium.lsm'))

plot_signal_image(
    hyperspectral_signal,
    axis=0,
    title='Hyperspectral image',
    xlabel='wavelength (nm)',
)

# %%
# Calculate phasor coordinates at the first harmonic and filter out
# pixels with low intensities:

mean, real, imag = phasor_from_signal(hyperspectral_signal, axis=0)
_, real, imag = phasor_threshold(mean, real, imag, mean_min=1)

# %%
# Plot the phasor coordinates as a two-dimensional histogram:

from phasorpy.plot import plot_phasor

plot_phasor(real, imag, allquadrants=True, title='Hyperspectral phasor plot')

# %%
# Find clusters
# -------------
#
# Automatically find the two elliptical clusters in the phasor space using the
# :py:func:`phasorpy.cluster.phasor_cluster_gmm` function and plot them
# in distinct colors:

from phasorpy.cluster import phasor_cluster_gmm

center_real, center_imag, radius_minor, radius, angles = phasor_cluster_gmm(
    real, imag, clusters=2
)

phasorplot = PhasorPlot(allquadrants=True, title='Elliptical clusters')
phasorplot.hist2d(real, imag, cmap='Greys')
for i in range(len(cursors_real)):
    phasorplot.cursor(
        center_real[i],
        center_imag[i],
        radius=radius[i],
        radius_minor=radius_minor[i],
        color=CATEGORICAL[i],
        linestyle='-',
    )
phasorplot.show()

# %%
# Use the elliptic clusters to mask regions of interest in the phasor space:

from phasorpy.cursors import mask_from_elliptic_cursor

elliptic_masks = mask_from_elliptic_cursor(
    real,
    imag,
    center_real,
    center_imag,
    radius=radius,
    radius_minor=radius_minor,
)

# %%
# Plot a pseudo-color image, composited from the elliptic cursor masks and
# the mean intensity image:

pseudo_color_image = pseudo_color(*elliptic_masks, intensity=mean)

plot_image(
    pseudo_color_image, title='Pseudo-color image from elliptic cursors'
)

# %%
# Appendix
# --------
#
# Print information about Python interpreter and installed packages:

print(phasorpy.versions())

# %%
# sphinx_gallery_thumbnail_number = -8
# mypy: allow-untyped-defs, allow-untyped-calls
# isort: skip_file
```

Btw, I had to switch the order of `radius_minor` and `radius`. Any idea why?

src/phasorpy/cluster.py
cgohlke: `Defaults to 2`

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Parameters
+    ----------
+    real : array_like
+        Real component of phasor coordinates.
+    imag : array_like
+        Imaginary component of phasor coordinates.
+    sigma: float, default = 2.0
+        Scaling factor for the radii of major and minor axes.
+        Default to 2, which corresponds to the scaling of eigenvalues for a
```
cgohlke, line 55:
```suggestion
        Defaults to 2, which corresponds to the scaling of eigenvalues for a
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,180 @@
...
+    ...     phasor_cluster_gmm(real, imag, clusters=2)
+    ... )
+
+    """
+    real = numpy.asarray(real, dtype=numpy.float64)
+    imag = numpy.asarray(imag, dtype=numpy.float64)
+
+    if real.shape != imag.shape:
+        raise ValueError(f'{real.shape=} != {imag.shape=}')
+
```
cgohlke, line 123: I don't think this is necessary.

```suggestion
```

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,180 @@
...
+    coords = numpy.stack((real, imag), axis=-1).reshape(-1, 2)
+
+    # coords: NDArray[numpy.float64] = numpy.column_stack(
+    #     (real.ravel(), imag.ravel()) # type: ignore[union-attr]
+    # )
+
```
cgohlke, line 129: Remove
```suggestion
```

tests/test_cluster.py
```diff
@@ -0,0 +1,93 @@
+"""Tests for the phasorpy.cluster module."""
+
+import numpy
+import pytest
+from numpy.testing import (
+    assert_allclose,
+    assert_almost_equal,
+    assert_array_equal,
+)
```
cgohlke, line 9: Currently not used

src/phasorpy/cluster.py
R-Blanco802: I believe that the pre-commit is erasing the blank line.
Because I had it, until I ran the `python -m pre_commit run --all-files`

cgohlke: Strange, not on my system.

R-Blanco802:
> elliptic_masks = mask_from_elliptic_cursor(
>     real,
>     imag,
>     center_real,
>     center_imag,
>     radius=radius,
>     radius_minor=radius_minor,
> )


> Btw, I had to switch the order of `radius_minor` and `radius`. Any idea why?

Yes, you should add the angles to the mask_from_elliptic_cursor, like:

`for i in range(len(center_real)):`
 `    plot.cursor(`
`         center_real[i],`
`         center_imag[i],`
`         radius=radius[i],`
`         radius_minor=radius_minor[i],`
`         angle = angles[i],`
`         color=CATEGORICAL[i],`
`         linestyle='-',`
`      )`
`plot.show()`

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,172 @@
...
+    Raises
+    ------
+    ValueError
+        If the array shapes of `real` and `imag` do not match.
+        If `clusters` is not a positive integer.
+
+    See Also
+    --------
+    :ref:`sphx_glr_tutorials_api_phasorpy_cluster.py`
```
cgohlke, line 92: Remove for now:

```suggestion
```

cgohlke:
Hi @bruno-pannunzio : Can you also review this PR? I think it is ready to be merged. It is a very nice addition to the library. The examples, tests, and tutorial can be improved, and a dedicated tutorial be added in followup PRs.

src/phasorpy/cluster.py
```diff
@@ -0,0 +1,169 @@
...
+    radius_minor : tuple of float
+        Minor radii of ellipses.
+    angle : tuple of float
+        Rotation angles of major axes in radians, within range [0, pi].
+
+    Raises
+    ------
+    ValueError
+        If the array shapes of `real` and `imag` do not match.
+        If `clusters` is not a positive integer.
```
bruno-pannunzio, line 88: Maybe I missed something but I don't see this errors raised explicitly within the function. Although I see this conditions will raise errors by inner numpy or scipy's functions. Is it okay to still leave this section anyhow? @cgohlke

cgohlke: You have a point. However, it's still correct and adds value to the documentation I think.  Here's the relevant section from the numpydoc style-guide: https://numpydoc.readthedocs.io/en/latest/format.html#raises

bruno-pannunzio (approved):
I believe this is ready to be merged, @cgohlke. Many thanks, @R-Blanco802, for your hard work in adding this functionality and integrating it into the library!

We can think later how to improve the tests and documentation but this is excellent work and it's a great addition to the library.

PR #229, opened 2025-05-28
Read first channel from files by default

cgohlke:
## Description

This PR changes the `signal_from_ptu`, `phasor_from_ifli`, `phasor_from_flimlabs_json`, and `signal_from_flimlabs_json` functions to return only the first channel by default (`channel=0`). Size-one dimensions are no longer returned by default (`keepdims=False`). This is a breaking change.

These functions currently return all channels stored in files and sometimes keep size-one dimensions in the output array. This regularly caused issues, often unnoticed, since PhasorPy functions require single channels (for example, all plotting functions) or special handling of multiple channels (for example, `phasor_calibrate`).

## Checklist

### DIFF
--- a/src/phasorpy/_io.py
+++ b/src/phasorpy/_io.py
@@ -822,7 +822,7 @@ def phasor_from_ifli(
     filename: str | PathLike[Any],
     /,
     *,
-    channel: int | None = None,
+    channel: int | None = 0,
     harmonic: int | Sequence[int] | Literal['all', 'any'] | str | None = None,
     **kwargs: Any,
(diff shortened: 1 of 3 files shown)


bruno-pannunzio (approved):
Hi @cgohlke, I totally agree with these changes. I think by far in most cases of use I had been specifying the first channel and the 'keepdims=False', so this makes a lot of sense for most of the people I guess. Also, in this line, I always use the average of all frames to improve the signal to noise ratio. I don't know if this is the case for most people, but maybe setting 'frame=-1' by default makes sense, but maybe not.

PR #235, opened 2025-06-03
Reorganize io module

cgohlke:
## Description

This PR reorganizes the `io` module by moving the implementation from `_io.py` into several private submodules in an `io` folder. The tests are reorganized accordingly. I/O helper functions are moved into the `phasorpy._utils` module. The public API remains unchanged.

## Checklist

- [ ] The pull request title and description are concise.
- [ ] Related issues are linked in the description.
- [ ] New dependencies are explained.
- [ ] The source code and documentation can be distributed under the [MIT license](https://www.phasorpy.org/docs/stable/license/).
- [ ] The source code adheres to [code standards](https://www.phasorpy.org/docs/stable/contributing/#code-standards).
- [ ] New classes, functions, and features are thoroughly [tested](https://www.phasorpy.org/docs/stable/contributing/#tests).
- [ ] New, user-facing classes, functions, and features are [documented](https://www.phasorpy.org/docs/stable/contributing/#documentation).
- [ ] New features are covered in tutorials.
- [ ] No files other than source code, documentation, and project settings are added to the repository.

PR #236, opened 2025-06-04
Sort clusters returned by phasor_cluster_gmm

cgohlke:
## Description

This PR fixes the order in which clusters are returned by the `phasor_cluster_gmm` function. Clusters are now returned in sorted order, by polar coordinates (default), phasor coordinates, or inverse ellipse area.

In #235, the `test_phasor_cluster_gmm_basic[2]` test started failing because the order of clusters returned by  `phasor_cluster_gmm` changed for unknown reason. 

## Checklist

### DIFF
--- a/src/phasorpy/cluster.py
+++ b/src/phasorpy/cluster.py
@@ -46,8 +46,6 @@ def phasor_cluster_gmm(
     extract the parameters of ellipses that represent each cluster according
     to [1]_.
 
-    Clusters are returned in descending order of their elliptic area.
-
     Parameters
     ----------
     real : array_like
--- a/src/phasorpy/cluster.py
+++ b/src/phasorpy/cluster.py
@@ -45,6 +45,8 @@ def phasor_cluster_gmm(
     extract the parameters of ellipses that represent each cluster according
     to [1]_.
 
+    Clusters are returned in descending order of their elliptic area.
+
     Parameters
     ----------
     real : array_like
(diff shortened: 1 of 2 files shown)


bruno-pannunzio:
Hi @cgohlke, I believe it is a good idea to make more consistent the order returned by the clusters. I was thinking of the case where you want to compare multiple samples and for example the area of the clusters changes between samples for some reason, then the order of the samples will be swapped and may complicate the analysis of heterogeneous datasets.

Maybe there are alternative ways to order the clusters that can be more consiste even for datasets with variable data. I am not sure which way can be more robust for this cases. Ordering based on the position of the center of the cluster maybe? It can be the coordinates or the phase/modulation? I think that the order of the clusters in the space is less prone to change in between samples of different phasor distributions. But also maybe there is another better option. Nevertheless, ordering by the area is much better than not ordering at all.

cgohlke:
The primary issue is to return clusters in a reproducible way in order to fix the test error and keep the tutorial output stable between runs.

Users can easily sort clusters however they like. It would also be easy to add an option for pre-defined sort orders, for example `{'phasor', 'polar', 'area'}`. For phasor applications (both lifetime and spectral), sorting by polar coordinates is probably more intuitive than phasor coordinates. Still, for datasets with variable data, the user must verify that clusters are matching if that is important. The situation is similar to working with segmentation labels.

I don't understand why the order of clusters changed in #235 but I had already noticed that issue some time ago on my system.

bruno-pannunzio (approved):
This is great @cgohlke, thank you very much for implementing different sorting option for the order returned of the clusters! This looks perfect to me.

Issue #206, opened 2025-03-20
Reorganize package namespace


cgohlke:
The purpose of this issue is to discuss if or how to reorganize the phasorpy package namespace, internal organization, and documentation before the next release (which should be the last release with large breaking changes).

The main arguments for reorganizing the package are that many imports feel clumsy (for example, `from phasorpy.phasor import phasor_from_signal`) and some modules grew too large (for example, `phasor.py` and `io.py`).

I propose the following reorganization:

1. Split the `phasor.py` module into `phasor`, `lifetime`, `fret`, and `filter`.

2. Export all public attributes from the `phasor`, `lifetime`, `fret`, `filter`, `components` `cursors`, `cluster`, and `color` modules from the `phasorpy` namespace. Above import will become `from phasorpy import phasor_from_signal`. 

3. Make the `phasor`, `lifetime`, `fret`, `filter`, `components` `cursors`, and `cluster` modules private, for example, prefix the names with an underscore. Should the `color` module be made private too?

4. Move the `io` module into a directory and split the implementation into separate, private files as necessary. All functions will be exported from the `phasorpy.io` namespace. The reasons for not moving io functions to the main namespace are that the `io` namespace is well established among other packages and functions are found faster.

5. Leave the `plot` module as is for now. It could be later moved into a directory and split up. The reason for not moving functions and classes from the plot module into the main namespace is that they are specific to the matplotlib plotting library.

6. Leave the `cli`, `datasets`, `utils`, and `version` modules as is. These are not directly related to PhasorPy's core functionality.

7. Rename or move some functions to make them consistent in the global namespace:

      - `n_fractions_from_phasor` -> `phasor_component_?`
      - `two_fractions_from_phasor` -> `phasor_component_?`
      - `graphical_component_analysis` -> `phasor_component_graphical`
      - `mask_from_circular_cursor` -> `phasor_mask_circular`
      - `mask_from_elliptic_cursor` -> `phasor_mask_elliptic`
      - `mask_from_polar_cursor` -> `phasor_mask_polar`
      - `utils.spectral_vector_denoise` -> `filter.spectral_vector_denoise`
      - `cursors.pseudo_color` -> `color.pseudo_color` or `image_overlay` or `image_blend`
      - `CATEGORICAL` -> ?
      - `SRGB_SPECTRUM` -> ?
      - `float2int` -> ?
      - `wavelength2rgb` -> ?

8. Reorganize tests reflecting the new package structure.

[...]

PR #240, opened 2025-06-09
Rename functions in components module

cgohlke:
## Description

This PR proposes to rename two functions in the `phasorpy.components` module as suggested/discussed many times, for example in #206, #215, and #106:

- `two_fractions_from_phasor` -> `phasor_component_fraction`
- `graphical_component_analysis` -> `phasor_component_graphical`

This makes function naming more consistent with other parts of the library, which is especially important when moving to a flat namespace. The `two_fractions_from_phasor` name was misleading since the function returns only a single fraction.

The functions were also changed to use `component_real` and `component_imag` parameter names throughout.

If there are better names for the functions/parameter names in the components module, let's discuss and settle this in this PR.

## Checklist

### DIFF
--- a/src/phasorpy/components.py
+++ b/src/phasorpy/components.py
@@ -3,28 +3,29 @@
 The ``phasorpy.components`` module provides functions to:
 
 - calculate fractions of two known components by projecting onto the
-  line between the components (:py:func:`two_fractions_from_phasor`)
+  line between the components (:py:func:`phasor_component_fraction`)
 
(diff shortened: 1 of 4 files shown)


bruno-pannunzio (approved):
Hi @cgohlke, I like the new names for the functions in the components module. Although the `phasor_component_fraction` may not be directly intuitive that it is between two components , I don't see a better alternative, and as you pointed out, it just returns the fraction of one component. So I think this name is a good compromise all considered.

PR #297, opened 2025-08-29
Add signal_filter_ncpca function

cgohlke:
## Description

This PR adds the `signal_filter_ncpca` function to the `phasorpy.filter` module. It implements the **noise-corrected Principal Component Analysis** (NC-PCA) denoising method described in:

> Soltani S, Paulson J, Fong E, Mumenthaler, S, and Arman A.
> Denoising of fluorescence lifetime imaging data via principal component analysis.
> *Preprint*, (2025). https://doi.org/10.21203/rs.3.rs-7143126/v1

The implementation is based on `sklearn.decomposition.PCA`. If necessary, this could be re-implemented using numpy only.

I manually verified that the results obtained for the "Embryo.tif" dataset match the results from [reference implementation](https://zenodo.org/records/14895224) within ~1e-10. Unfortunately the authors did not publish their synthetic datasets.

## Checklist

### DIFF
--- a/src/phasorpy/filter.py
+++ b/src/phasorpy/filter.py
@@ -11,9 +11,10 @@ The ``phasorpy.filter`` module provides functions to filter
 
 - signals:
 
+  - :py:func:`signal_filter_ncpca`
+    (noise-corrected principal component analysis)
   - :py:func:`signal_filter_svd` (spectral vector denoise)
   - :py:func:`signal_filter_median` (not implemented yet)
-  - :py:func:`signal_filter_pca` (not implemented yet)
 
 """
 
@@ -25,7 +26,7 @@ __all__ = [
     'phasor_filter_pawflim',
     'phasor_threshold',
     # 'signal_filter_median',
-    # 'signal_filter_pca',
+    'signal_filter_ncpca',
     'signal_filter_svd',
 ]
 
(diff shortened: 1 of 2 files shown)


src/phasorpy/filter.py
```diff
@@ -823,3 +824,139 @@ def signal_filter_svd(
...
+    Raises
+    ------
+    ValueError
+        If `dtype` is not a floating point type.
+        If `signal` has fewer than 3 samples along specified axis.
+        If `n_components` is invalid for the data size.
+
+    References
+    ----------
+    .. [3] Soltani S, Paulson J, Fong E, Mumenthaler, S, and Arman A.
```
bruno-pannunzio, line 884: I think the last author surname is Armani

bruno-pannunzio (approved):
Hi @cgohlke, this looks very promising. I will test this further with different datasets to compare it with other methods for filtering, just for curiosity. Maybe at some point we can have a tutorial comparing methods for filtering, just for demonstration purposes, we have something in mind for the binlet filtering also. 

I think it's okay for now as you proposed to use sklearn. I havent tested execution times for large signals, but if this becomes a problem we can think of using pure numpy as you proposed. It would be nice to get feedback from users on this.

But overall, I think everything looks ok here, so thank you for adding this functionality. I think everything works, I will try with some real data and open an issue if something is not as expected.

cgohlke:
Hi @bruno-pannunzio 

The algorithm/implementation of this method is quite simple. The runtime doesn't seem to be a problem so far. I don't think an implementation in pure numpy will be faster, just not dependent on sklearn. The reference implementation calculates and verifies sklearn and numpy implementations by default, which is why it is notably slower.

The documentation already mentioned some things to be aware of:

https://github.com/phasorpy/phasorpy/blob/d6f1729f6a6b2e84dbf92fbd92414bdbac23a46b/src/phasorpy/filter.py#L872-L876

https://github.com/phasorpy/phasorpy/blob/d6f1729f6a6b2e84dbf92fbd92414bdbac23a46b/src/phasorpy/filter.py#L892-L897

Another potential issue is that PCA doesn't work with NaN/Inf input. This method replaces NaN/Inf with zero, which I'm not sure will affect PCA projection?

PR #340, opened 2026-05-21
Add signal_from_brighteyes_mcs function

mattia-donato:
## Description

Adds an HDF5 reader for BrightEyes-MCS data.

The new reader supports HDF5 files containing raw histogram data in data with shape (repetition, z, y, x, time, channel). It selects a repetition, z-plane, and channel, then returns a YXH signal suitable for phasor analysis.

The reader also supports loading calibration traces from calibration/data/ref_common_delay_realigned for REF and calibration/data/irf_common_delay_realigned for IRF. These calibration datasets are expected to have shape (time, channel) and are returned as a virtual 1 x 1 image with shape (1, 1, time) for compatibility with calibration routines.

Adds tests for data, REF, and IRF loading, and registers the new HDF5 reader in phasorpy.io.

## Checklist


cgohlke:
Hi @mattia-donato : Thanks for proposing this! I have not reviewed in detail yet but have a couple of general questions. 

- Is there a formal specification of the BrightEyes-MCS h5 file format that we could reference? 
- Is the file format stable?
- Are any public data files available for testing and reference? 
- Is the format related to [Photon-HDF5](https://photon-hdf5.github.io/) (I guess not)? 
- Rather than implementing the file format reader in this repo, would it make sense to move the reader details to a separate package? 

I see you have started [BrightEyes-MCS-File](https://github.com/VicidominiLab/BrightEyes-MCS-File), which is licensed under GPL. On the other hand, if the file format is stable, the implementation seems small enough to keep it inline here.

cgohlke:
Hi @mattia-donato, 

I would like to **discuss** changing the following:

1. remove `phasor_from_h5` convenience function. The files do not store phasor coordinates.
2. rename `signal_from_h5` to `signal_from_mcsh5` or `signal_from_brighteyes_mcs`. That avoids conflict with other HDF5 based formats.
3. rename `repetition` parameter to `time`, `'R'` dimension to `'T'`. That matches other file formats.
4. rename `z` parameter to `depth`. It is more descriptive.
5. remove `dtype` parameter. Users can call .astype() on the returned array.
6. replace the five parameters `data_dataset`, `reference_dataset`, `irf_dataset`, `reference`, and `irf` with a single `dataset: str | None = None`.

    When None (the default), the function tries 'data' or scans the HDF5 file for a single 6D dataset. The strings 'reference' and 'irf' act as reserved aliases that resolve to the default calibration dataset paths and set the appropriate metadata. Any other string is treated as a literal HDF5 path, with the branch (data vs. calibration) determined by the dataset's dimensionality (6D -> image data, 2D -> calibration/reference).
  
    This simplifies common call sites (`reference=True` becomes `dataset='reference'`, `irf=True` becomes `dataset='irf'`) while still allowing custom dataset paths without extra flags.
    
    Possible ambiguities to flag:
    
    - Magic string collision. If an HDF5 file contains a dataset literally named reference or irf at the root, there is no way to address it by that name; the function would intercept those strings as aliases. Users would need to prefix with a slash (e.g. dataset='/reference') or rename — unless the implementation explicitly reserves only unqualified single-word aliases.
    - IRF vs. reference metadata for custom paths. When a custom path is passed to a 2D dataset, the function cannot distinguish an IRF from a reference histogram. The irf metadata key would be absent from attrs, which is a behavioral difference from the current irf=True path.
    - Auto-detection failure modes. With dataset=None, files containing zero or more than one 6D dataset raise a ValueError. This is a stricter default than the current behavior (which always tries 'data'), and may affect users whose files follow the standard naming but also contain other 6D datasets.

In any case, we would need a real file for testing (and the I/O tutorial if possible), **ideally** a small (few MB), public (permissively licensed), real measurement (mentioned in a publication) file containing data, reference, irf, and metadata.

cgohlke:
Hi @mattia-donato : would you be able to share a BrightEyes-MCS files with FLIM and calibration data? I could not find such a file in your group's Zenodo records.

mattia-donato:
Hi @cgohlke, thank you for your comments.

Your feedback triggered an internal discussion about updating the HDF5 schema.

We decided that, for the moment, the schema of the “raw” HDF5 files produced by BrightEyes-MCS will not be upgraded. An example raw dataset is already available here: https://doi.org/10.5281/zenodo.11284051

Please note, however, that these “raw” data are not directly usable for phasor analysis. They require an additional calibration step. By “calibration” I do not mean phasor calibration, but rather a time-alignment correction required by the specific DFD implementation in the firmware. Therefore, PhasorPy and napari-phasors will mainly use the calibrated files.

We decided to revise the HDF5 schema of the “calibrated” and “output” files to make it more self-descriptive. 

I will contact you again very soon and publish the updated schema, together with example files, on Zenodo or a similar platform. For now, thank you again for your patience and for the useful feedback.

cgohlke:
See also https://github.com/napari-phasors/napari-phasors/pull/280

cgohlke:
Hi @mattia-donato : any updates on the revised schema or sample files?

mattia-donato:
Hi @cgohlke 

Thank you for your patience!

I have some good news: I finally uploaded an example dataset to Zenodo. You can find it here:

https://zenodo.org/records/21158280?token=eyJhbGciOiJIUzUxMiJ9.eyJpZCI6Ijk5ZTgzODJlLWY0MzItNDEwZi04ZTBjLWYwNjJlN2Q1NGQ5YSIsImRhdGEiOnt9LCJyYW5kb20iOiJiMjQ5OTE3ZTA3Mzg5NzYwNjE1MDVmOTE3NWFhN2U0YyJ9.kqQvfaAynyeukD3ACDrO75fbZtDZryrN4-E53DkZdm_Ev1yDLU-EUQDYYaScIaW7Lb7tEPzYuz17yKL8JMHu9Q

The dataset includes a couple of files acquired from a Convallaria sample. Each file also contains the internal calibration reference measurement.

The files above should follow the BrightEyes H5 scheme, version 0.0.6, which is available in the repository here:

https://github.com/VicidominiLab/BrightEyes-MCS-File/blob/main/docs/Brighteyes%20H5%20SCHEME.txt

The screenshots below were taken with napari, using the napari-phasor plugin and phasorpy, after updating my forks of both napari-phasor and phasorpy. The code is still not very clean and was heavily AI-assisted, but at least it allowed us to visualize the data successfully. :-)

Ciao,
Mattia

image

image

cgohlke:
Hi @mattia-donato : thanks for sharing the sample files and updating the implementation for the new schema!

Any thoughts on my previous comments? https://github.com/phasorpy/phasorpy/pull/340#issuecomment-4510538954

Given the extent of this implementation, the size of the test files, and the fact that the schema might get updated, I think it might be better to put the implementation in a separate, permissively licensed (like MIT or BSD) library and provide a thin wrapper function in PhasorPy. This has been done with PTU and LIF files for example. What do you think?

cgohlke:
Hi @mattia-donato : Thank you! The wrapper approach looks much more maintainable and flexible.

Before we continue, I wanted to ask whether you would consider re-licensing brighteyes-mcs-file under a more permissive license, such as MIT, BSD, or Apache 2.0. The concern is that using (calling or linking to) a GPL-licensed library requires PhasorPy and all dependent packages (like napari-phasors) to be distributed under the GPL as well.

mattia-donato:
Thanks for the suggestion.

I created [BrightEyes-MCS-Reader](https://github.com/VicidominiLab/BrightEyes-MCS-Reader) with a BSD-3-Clause license.

Since the previous name did not make much sense anymore, I also moved **BrightEyes-MCS-File** to [BrightEyes-MCS-DataPrep](https://github.com/VicidominiLab/BrightEyes-MCS-DataPrep).

I also updated the PyPI packages accordingly, so users can now install the packages with:

``pip install brighteyes-mcs-dataprep``

and:

``pip install brighteyes-mcs-reader``

In principle, most users should only need **BrightEyes-MCS-Reader**.

**BrightEyes-MCS-DataPrep** is mainly for preparing/converting data, and the idea is that it will eventually be integrated directly into the **BrightEyes-MCS** GUI.

cgohlke:
Hi @mattia-donato : that's good news. Thank you.

Before enabling GitHub Actions on this PR, can you please address the following:

- https://github.com/VicidominiLab/BrightEyes-MCS-Reader does not seem to exist yet or is private (I get a 404).

- I think the implementation and tests are now small enough to move them into the `_others.py` and `test_others.py` modules. No need for `_mcs_h5.py` and `test_h5.py`. What do you think?

- Undo the changes to `test__utils.py`. They are not necessary and currently have a merge conflict.

- Update your branch from main.

- If you have not already done so, run `pre-commit` and `mypy` on your branch (https://www.phasorpy.org/docs/stable/contributing/#code-standards) and resolve any issues.

mattia-donato:
Sorry I put private visibility by mistake now is public https://github.com/VicidominiLab/BrightEyes-MCS-Reader  . For the rest I will try to check tomorrow.

src/phasorpy/plot/_phasorplot.py
```diff
@@ -553,7 +553,7 @@ def components(
         center_re, center_im = numpy.average(
             numpy.vstack([real, imag]),
             axis=-1,
-            weights=fraction,  # type: ignore[arg-type]
+            weights=fraction,
```
cgohlke:
```suggestion
            weights=fraction,  # type: ignore[arg-type]
```

src/phasorpy/io/_other.py
```diff
@@ -37,6 +38,88 @@
...
+    filename: str | PathLike[Any],
+    /,
+    *,
+    dataset: str | None = None,
+    time: int = 0,
+    depth: int = 0,
+    channel: int | Literal['sum'] = 0,
+) -> DataArray:
+    """Return histogram image from BrightEyes-MCS HDF5 file.
+
```
cgohlke, line 51: Add a sentence about BrightEyes-MCS HDF5 files: what kind of data do they contain; which software writes the files.

src/phasorpy/io/_other.py
```diff
@@ -37,6 +38,88 @@
...
+        Dataset to read. By default, read the BrightEyes-MCS default image
+        dataset. The reserved aliases ``"reference"`` and ``"irf"`` read the
+        default calibration reference and IRF traces. Other strings are
+        interpreted as literal HDF5 dataset paths.
+    time : int, optional, default: 0
+        Index along the BrightEyes-MCS repetition/time axis.
+    depth : int, optional, default: 0
+        Index along the BrightEyes-MCS z/depth axis.
+    channel : int or 'sum', optional, default: 0
+        Channel index to read. If ``"sum"``, integrate the channel axis.
```
cgohlke, line 66: So it's currently not possible to read/return the whole dataset as one array. Is the channel sum calculated after reading or read from the file? Should it be documented that the raw SPAD channels can be read from `dataset='/raw/spad'`? `channel=1` will always fail for the default dataset, no?

src/phasorpy/io/_other.py
```diff
@@ -37,6 +38,88 @@
...
+
+    Returns
+    -------
+    xarray.DataArray
+        Histogram image with dimensions such as ``('Y', 'X', 'H')``.
```
cgohlke, line 71: Be more specific about the returned array: shape, dtype, important attributes. See the docstrings of similar functions. Something like

```
TCSPC histogram with axes codes 'YXH', type `float32`, `float64` or `uint16` (depending on dataset), and selected metadata:

- coords['H']: delay times of histogram bins in ns.
- attrs['frequency']: repetition frequency in MHz.
- attrs['reference_lifetime_ns']: reference lifetime in ns.
```

tests/io/test_other.py
```diff
@@ -23,6 +28,136 @@
...
+        ref.attrs['trace_kind'] = 'sum_reference_trace'
+        ref.attrs['time_axis_path'] = '/output/sum_ref_001/axes/time_ns'
+
+    return output, reference
+
+
+def test_signal_from_brighteyes_mcs(tmp_path: Path) -> None:
+    """Test wrapping a BrightEyes-MCS signal as xarray."""
+    pytest.importorskip('brighteyes_mcs_reader')
+    pytest.importorskip('xarray')
```
cgohlke:
```suggestion
```

tests/io/test_other.py
```diff
@@ -23,6 +28,136 @@
...
+    )
+    assert signal.attrs['frequency'] == 80.0
+    assert signal.attrs['reference_lifetime_ns'] == 2.7
+    assert signal.attrs['h5_dataset'] == '/output/sum_001/products/spad'
+
+
+def test_signal_from_brighteyes_mcs_reference(tmp_path: Path) -> None:
+    """Test reading a default BrightEyes-MCS reference trace."""
+    pytest.importorskip('brighteyes_mcs_reader')
+    pytest.importorskip('xarray')
```
cgohlke:
```suggestion
```

cgohlke:
Hi @mattia-donato : Thank you again! It looks like this is basically ready. All tests pass. I will push some of the suggested minor changes and merge. Improvements and remaining review issues can be addressed in a followup PR.

tests/io/test_other.py
```diff
@@ -23,6 +28,132 @@
 )
 
 
+def _write_brighteyes_mcs_h5(
+    filename: str | Path,
+) -> tuple[NDArray[Any], NDArray[Any]]:
+    h5py = pytest.importorskip('h5py')
```
cgohlke:
```suggestion
    import h5py
```

tests/io/test_other.py
```diff
@@ -23,6 +28,132 @@
 )
 
 
+def _write_brighteyes_mcs_h5(
+    filename: str | Path,
+) -> tuple[NDArray[Any], NDArray[Any]]:
+    import h5py
```
cgohlke, line 34:
```suggestion
    import h5py

```

cgohlke (approved):

